PR #140, opened 2014-08-13
Pricing Attachment Observer

chrissrogers:
##### Refactors
- `recurly.pricing.attach` now acts as a factory for the new `Attachment` local object prototype.
- `recurly.pricing.attach` used to intertwine heavily with the `Pricing` prototype. This dependency is now injected via the `Attachment` initializer.
##### New Features
- `Attachment` triggers an observer on the passed element. Changes to the DOM subtree will initialize a dirty check. If the subtree's recurly elements are not as they were before the change, `Attachment` will refresh its bindings.
- This eliminates the integrators' need to understand when the pricing attachment needs to be rebound, and makes for a much more efficient rebinding process.
- The present way of refreshing bindings is rife with opportunity for memory leaks. This controls the rebinding much more closely, reducing the risk of leaks.
##### TODO
- Tests

/cc @cbarton @bhelx


chrissrogers:
🔨  **obsolete!**

- the better way to do this is with a React component

Issue #176, opened 2015-01-26
Coupon code callback


merauluka:
We are updating our code from v1 to v3 of the Recurly.js API. We have everything working so far except the coupon code field.

**Is there a way to return some meaningful value if the code is invalid?** At the moment we have the pricing information calculating, but the coupon code appears to only return data when the code is valid. *If the code is invalid it would be great to have some kind of message we could pass on to the user.*

We would like to mimic the functionality that used to appear in Recurly.js's buildSubscriptionForm form building call.

chrissrogers:
> Is there a way to return some meaningful value if the code is invalid?

If you are calling the pricing API directly, the coupon method will throw an error with code 'not-found' for coupons which do not apply to the selected plan.

If you are attaching pricing to a form, then bad coupon codes are simply ignored [1]. I could see it being useful for an event to fire on the pricing instance if this occurs, though.

[1]: https://github.com/recurly/recurly-js/blob/master/lib/recurly/pricing/attach.js#L126-L129

merauluka:
I am attaching pricing to my form. I ended up calling the pricing API directly when the coupon field is updated because I wanted to tell my users the status of their coupon code. But it would be cool to be able to do it all at once instead.

chrissrogers:
Agreed. I think that instead of silencing not-found coupon errors, we can instead emit an event. `invalid.coupon` or something similar.

thbar:
There is more things lacking with coupon JS API, actually. See [here](#225) for instance.

I would have assumed that the `done(price)` callback would have add things like coupon validity (yes/no),  period of applicability, a label.

I think the JS library is a bit lagging with functionalities around this area! (sure, I could issue an extra API call myself, but why do that when the Recurly.js library already does it but does not expose the information?).

Thanks for considering this!

PR #177, opened 2015-01-27
Fixes issue with pricing coupon and plan mismatches

### LINKED ISSUE #175: Validating plan-restricted coupons
When creating a code that is restricted to certain plans, Recurly.js doesn't appear to be updating the totals correctly when the coupon is set and then the plan is changed. The totals update, but the coupon is applied regardless of the plan restriction.

### Steps to reproduce:

1. Select the plan (we are using radio inputs).
2. Enter the coupon code that works with that plan.
3. Change the plan.

At this point the coupon's discount is applied even though it is restricted to only one plan. I can change the plan to any of our listed subscription plans and it still calculates the discount.

### Inverse issue:

1. Select the plan (one that isn't allowed on the coupon)
2. Enter the coupon code.
3. Change the plan to the one that is allowed on that coupon.

The discount remains at $0 even when the correct plan is chosen. No matter which plan is chosen, the discount remains at $0.

### Coupon settings

Just in case it matters, the coupon we setup has the following options:

1. 25% discount
2. Single use
3. Unlimited redemptions
4. Never expires
5. Only affects one plan (which is annual)

Is this on anyone's radar? Or is there a way to work around it?

chrissrogers:
- Calls to pricing.coupon will no longer abort if the coupon code hasn't changed. This accomodates plan changes that may nullify a coupon that may wok on other plans
- Corrects not-found error code check in pricing.coupon
- pricing.plan calls will now make pricing.coupon calls if a coupon is set on the instance
- Fixes #175

### DIFF
--- a/lib/recurly/pricing/attach.js
+++ b/lib/recurly/pricing/attach.js
@@ -24,7 +24,7 @@ exports.attach = function (el) {
 
   if (this.attach.detatch) this.attach.detatch();
 
-  self.on('change', update);
+  this.on('change', update);
 
   each(el.querySelectorAll('[data-recurly]'), function (elem) {
     // 'zip' -> 'postal_code'
@@ -53,11 +53,11 @@ exports.attach = function (el) {
       pricing = pricing.currency(dom.value(elems.currency));
     }
 
-    if (target('addon') && elems.addon) {
+    if (elems.addon && target('addon')) {
       addons();
     }
 
-    if (target('coupon') && elems.coupon) {
+    if (elems.coupon && (target('coupon') || target('plan'))) {
       pricing = pricing.coupon(dom.value(elems.coupon)).then(null, ignoreBadCoupons);
     }
 
--- a/lib/recurly/pricing/index.js
+++ b/lib/recurly/pricing/index.js
@@ -170,9 +170,18 @@ Pricing.prototype.plan = function (planCode, meta, done) {
         self.currency(keys(plan.price)[0]);
       }
 
-      debug('set.plan');
-      self.emit('set.plan', plan);
-      resolve(plan);
+      // If we have a coupon, it must be reapplied
+      if (self.items.coupon) {
+        self.coupon(self.items.coupon.code).then(finish, finish);
+      } else {
+        finish();
+      }
+
+      function finish () {
+        debug('set.plan');
+        self.emit('set.plan', plan);
+        resolve(plan);
+      }
     });
   }, this).nodeify(done);
 };
@@ -203,8 +212,8 @@ Pricing.prototype.addon = function (addonCode, meta, done) {
     var planAddon = findAddon(self.items.plan.addons, addonCode);
     if (!planAddon) {
       return reject(errors('invalid-addon', {
-          planCode: self.items.plan.code
-        , addonCode: addonCode
+        planCode: self.items.plan.code,
+        addonCode: addonCode
       }));
     }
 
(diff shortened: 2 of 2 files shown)


lib/recurly/pricing/attach.js
```diff
@@ -53,11 +53,11 @@ exports.attach = function (el) {
       pricing = pricing.currency(dom.value(elems.currency));
     }
 
-    if (target('addon') && elems.addon) {
+    if (elems.addon && target('addon')) {
       addons();
     }
 
-    if (target('coupon') && elems.coupon) {
+    if (elems.coupon && (target('coupon') || target('plan'))) {
```
chrissrogers: I switched these around since the `target` method calls are a bit more expensive. Should speed things up marginally for forms not using addons or coupons.

lib/recurly/pricing/index.js
```diff
@@ -243,14 +252,18 @@ Pricing.prototype.coupon = function (couponCode, done) {
...
+    if (!couponCode) {
+      debug('unset.coupon');
+      delete self.items.coupon;
+      return resolve();
     }
-    if (!couponCode) return resolve();
 
     self.recurly.coupon({ plan: self.items.plan.code, coupon: couponCode }, function (err, coupon) {
-      if (err && err.code !== 'not_found') return reject(err);
+      if (err && err.code !== 'not-found') return reject(err);
```
chrissrogers: I'd like to follow up on this PR with some reconsideration of how we handle not-found coupons. Right now we ignore the error completely, but I think we should probably emit an event that merchants can listen to. See #176

lib/recurly/pricing/index.js
```diff
@@ -243,14 +252,18 @@ Pricing.prototype.coupon = function (couponCode, done) {
...
-      if (coupon.code === couponCode) return resolve(coupon);
-      else self.remove({ coupon: coupon.code });
+    if (coupon) self.remove({ coupon: coupon.code });
+
+    // TODO: Should this emit an 'unset' event?
+    // Need to consider other objects as well
+    if (!couponCode) {
+      debug('unset.coupon');
+      delete self.items.coupon;
+      return resolve();
```
chrissrogers: This adds a new class of events: `unset` -- I'd like to follow this PR with one adding `unset` events to the remaining objects.

lib/recurly/pricing/index.js
```diff
@@ -243,14 +252,18 @@ Pricing.prototype.coupon = function (couponCode, done) {
 
   return new PricingPromise(function (resolve, reject) {
     if (!self.items.plan) return reject(errors('missing-plan'));
-    if (coupon) {
-      if (coupon.code === couponCode) return resolve(coupon);
-      else self.remove({ coupon: coupon.code });
+    if (coupon) self.remove({ coupon: coupon.code });
```
chrissrogers: This ensures we always process coupon set calls, since plans may have changed between calls. I don't think we should be tracking plan change histories in the coupon method.

lib/recurly/pricing/index.js
```diff
@@ -170,9 +170,18 @@ Pricing.prototype.plan = function (planCode, meta, done) {
...
 
-      debug('set.plan');
-      self.emit('set.plan', plan);
-      resolve(plan);
+      // If we have a coupon, it must be reapplied
+      if (self.items.coupon) {
+        self.coupon(self.items.coupon.code).then(finish, finish);
+      } else {
+        finish();
+      }
```
chrissrogers: Thus when we change plans, we reapply the potentially plan-dependent coupon.

We don't expose in the API which plans a coupon applies to since that may be sensitive information a merchant doesn't wish to expose publicly. The result is more API calls.

PR #178, opened 2015-01-27
Locks mocha-phantomjs to 3.5.2

chrissrogers:
- Fixes build
- Due to mocha-phantomjs incompatibility with our browser-based mocha
- Follow-up: update browser mocha

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "express": "3.x",
     "body-parser": "1.9.2",
     "jade": "1.x",
-    "mocha-phantomjs": "3.x",
+    "mocha-phantomjs": "3.5.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "express": "3.x",
     "body-parser": "1.9.2",
     "jade": "1.x",
-    "mocha-phantomjs": "3.x",
+    "mocha-phantomjs": "3.5.2",
```
bhelx: maybe while we are in here we should line up our colons vertically

bhelx:
:+1:

package.json
bhelx: jk

PR #179, opened 2015-01-27
Fixes validate.cardNumber throwing on falsey values

### LINKED ISSUE #161: Error thrown on empty credit card number
validate.cardNumber() will throw TypeError: Cannot read property 'length' of undefined

You can fix this by changing:

https://github.com/recurly/recurly-js/blob/master/lib/util/parse-card.js#L10

to:

`return (number || '').toString().replace(/[-\s]/g, '');`

And someone should add a test at https://github.com/recurly/recurly-js/blob/f877db8f04e6585fed50748cff2fc8ea71d61869/test/unit/validate.test.js#L12

like:

```javascript
   it('should return false for a null cardNumber', function() {
      assert(false === recurly.validate.cardNumber());
    });
```

Sure, a form should validate that you have a credit card number... but if someone tries to validate the form before putting anything in the number... it should just be false ;)

chrissrogers:
- instead of parseCard returning false, it returns an empty string that cardNumber can parse
- Fixes #161

/cc @bhelx

### DIFF
--- a/lib/util/parse-card.js
+++ b/lib/util/parse-card.js
@@ -7,5 +7,5 @@
  */
 
 module.exports = function parseCard (number) {
-  return number && number.toString().replace(/[-\s]/g, '');
+  return (number || '').toString().replace(/[-\s]/g, '');
 };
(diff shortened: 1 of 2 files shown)


bhelx:
This failing the travis tests for any specific reason?

chrissrogers:
Sometimes travis falls down with the following

```
/bin/sh: 1: node_modules/.bin/duo: not found
```

presumably because the build tool `duo`'s binary hasn't been provided. It happens at random and I have yet to figure out why this happens. It's not replicable locally, either. I've always 'solved' it by restarting builds.

PR #186, opened 2015-02-26
Fix tax exempt for non usst

bhelx:
Taxes are being added to tax exempt plans when the type is not usst. I think this should fix it but need to test. Any ideas?

Approvers: @chrissrogers


chrissrogers:
Travis is misbehaving. I ran the specs locally and am satisfied.

PR #187, opened 2015-03-05
Allows cardType to run detections on partial numbers

chrissrogers:
##### Summary
- It is common to want to detect a card's type before the entire card is input.
- Adds a second boolean parameter to the method.
- Passing true to the `partial` parameter instructs the method to ignore the card length detections and rely solely on pattern detection.


bhelx:
:+1:

PR #188, opened 2015-03-10
Fixes issue where Pricing detachment fails on multiple instances

chrissrogers:
##### Summary

- This places the detachment method on the true instance, instead of storing on the attach method prototype
- When multiple pricing instances from the same `Pricing` prototype definition are attached to separate forms, the pricing `attach` method would look for detach on the `attach` prototype. This should instead reside strictly on the instance.


bhelx:
:+1:

Issue #193, opened 2015-03-20
the example of coupon code doesn' work from the 3.0.11 version


luarmr:
The next code extract from the examples in https://docs.recurly.com/js/#pricing-api
```
var pricing = recurly.Pricing();

pricing
  .plan('basic', { quantity: 1 })
  .coupon('coop')
  .catch(function (err) {
    // err.code
  })
  .done(function (price) {
    // price object as emitted by 'change' event
  });
```
The error code is not executing when the coupon is failing.

I go back to the previous version and i write this for alert the people have the same problem, because i didn't search inside the code at the moment. If i will try to collect more info.

chrissrogers:
You are correct that the .catch will not catch coupon not found errors. This is due to the line changed in your PR; however, the line should remain, as it is not our intention to throw an error when a coupon is not found, and instead to fail silently while removing all coupons from the pricing object.

I think the best solution here is to emit an invalid coupon event, for which we have a discussion in #176

chrissrogers:
In versions previous to `3.0.11`, we were mistakenly emitting an error. This was fixed in #177

PR #194, opened 2015-03-20
Fix issue where Pricing dont fail when the coupon is invalid

luarmr:
The origin of this issue is here: https://github.com/recurly/recurly-js/issues/193
The fixed line was touched in the between 3.0.10 and 3.0.11 but not complete right.


luarmr:
Any chance to move this merge request.
I can't launch Travis with the correct config. Can you help with that?
Thanks!

chrissrogers:
This is actually the intended behavior. We only want to reject the promise if the error is of any type other than 'not-found'

chrissrogers:
Closing in favor of a solution to come from #176

PR #195, opened 2015-03-29
Fixes a test form input values not to use string 'undefined'

untidy-hair:
Hi there!

While I was working on something else with recurly.js, I found an improvement in test.
Looks like "when called with an HTMLFormElement" test creates a form with id="test-form", but the created form inputs have value 'undefined' IN STRING like so:

```html
...
<input type="text" data-recurly="cvv" value="undefined">
...
<input type="text" data-recurly="address1" value="undefined">    
<input type="text" data-recurly="address2" value="undefined">    
<input type="text" data-recurly="city" value="undefined"> 
....
```

This patch helps the form have correct intended values like so:

```html
...
<input type="text" data-recurly="cvv" value="">
...
<input type="text" data-recurly="address1" value="">    
<input type="text" data-recurly="address2" value="">    
<input type="text" data-recurly="city" value=""> 
....
```

Me personally, I think the fix is good enough for this case, but If you prefer other ways to fix this like defining a helper method instead of repeating  (values.keyName || ""), I could do that, too, let me know :)

### DIFF
--- a/test/unit/token.test.js
+++ b/test/unit/token.test.js
@@ -57,19 +57,19 @@ helpers.apiTest(function (requestMethod) {
         helpers.domTest(function (testbed, done) {
           testbed.insertAdjacentHTML('beforeend',
             ' <form id="test-form"> ' +
-            '   <input type="text" data-recurly="number" value="' + values.number + '"> ' +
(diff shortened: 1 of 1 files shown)


chrissrogers:
:+1: This is great, thanks!

PR #196, opened 2015-03-30
Add optional field validation

untidy-hair:
Hello,

At work, I needed additional field validations on client side so I created this PR :)
This PR gives us(merchants) a control about which additional required fields to be validated on browser.

**Changes:**
  Change token.validate() to validate 'cvv' (only when cvv exists) and merchant pre-defined required fields as well.

**How it works:**
  Merchant setups additional required fields with recurly.configure().
```javascript
recurly.configure({
  'publicKey': 'abcd1234-key',
  'extraRequiredFields': ['cvv', 'postal_code', 'country']
});
```
  When user submits form, token.validate() now checks if those required fields exist or not, too, through token().
  (Also, it checks 'cvv' only when cvv has a value.)

**Why this is helpful?**
 We use recurly.js at work and we wanted to check if "postal code", "cvv" and "country" are filled in subscription forms.
 We know they will be checked on server side on our setting but client validation is useful as follows:
 1) Improvement of user experience 
   When user submits form without "number" and "postal code", the form can show validation errors for both "number" and "postal code" at the same time, not just error for "number".
 2) The number of api calls can be reduced
   It can reduce the number of api calls in cases that "number" is filled but "postal code" is not filled.

**Caveat:**
 As you may have noticed, this depends on another PR #195. So if necessary, I will rebase this PR.

I may be missing something important, but this additional control would really help us. 
Thanks a lot :)

### DIFF
--- a/lib/recurly.js
+++ b/lib/recurly.js
@@ -103,6 +103,8 @@ Recurly.prototype.configure = function (options) {
     this.config.currency = options.currency;
   }
 
+  this.config.required = options.required || [];
+
   this.configured = true;
 };
 
--- a/lib/recurly/token.js
+++ b/lib/recurly/token.js
@@ -145,5 +145,15 @@ function validate (input) {
     errors.push('last_name');
   }
 
+  if (input.cvv && !this.validate.cvv(input.cvv)) {
+    errors.push('cvv');
+  }
+
+  each(this.config.required, function(field) {
+    if (!input[field] && ~index(fields, field)) {
+      errors.push(field);
+    }
+  });
+
   return errors;
 }
--- a/test/unit/configure.test.js
+++ b/test/unit/configure.test.js
@@ -17,6 +17,7 @@ describe('Recurly.configure', function () {
       , { invalid: 'parameter' }
       , { currency: 'USD' }
       , { currency: 'AUD', api: 'https://localhost' }
+      , { currency: 'USD', api: 'https://localhost', required: ['postal_code'] }
     ];
 
     it('throws', function () {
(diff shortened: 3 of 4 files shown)


lib/recurly/token.js
```diff
@@ -145,5 +145,15 @@ function validate (input) {
     errors.push('last_name');
   }
 
+  if (input.cvv && !this.validate.cvv(input.cvv)) {
+    errors.push('cvv');
+  }
```
chrissrogers: We actually don't require `cvv` in all cases, so this should definitely be optional.

lib/recurly.js
```diff
@@ -103,6 +103,8 @@ Recurly.prototype.configure = function (options) {
     this.config.currency = options.currency;
   }
 
+  this.config.extraRequiredFields = options.extraRequiredFields || [];
```
chrissrogers: I'd like to make this property a bit more succinct. Perhaps simply `required`. It's not always the case that form fields are being passed in, so I think a generic name is best.

chrissrogers:
This is a great contribution. Awesome test coverage, and I think it's a great feature to add as well.

Will need to be rebased.

lib/recurly/token.js
untidy-hair: If cvv is not there, the first condition returns false and 'cvv' will not be added to error. So I think the current implementation is good?

lib/recurly.js
untidy-hair: You've got it :)

untidy-hair:
@chrissrogers Thank you for the comment and the notes. Will be happy to rebase it when we make sure about the cvv part :)

lib/recurly/token.js
chrissrogers: Ah okay, I see now how this is first checking for input presence. Okay, that looks great!

chrissrogers:
@untidy-hair Once the property name has been changed and this is rebased, I'll be happy to run tests and merge.

untidy-hair:
Changed the property name and rebased :)
Thanks!

untidy-hair:
Cool, thanks!

PR #198, opened 2015-03-30
Tax Bug

seebees:
The math on $49 * .20 === $9.80 not $9.81


chrissrogers:
This is being caused by floating point math problems here: https://github.com/recurly/recurly-js/blob/master/lib/recurly/pricing/calculations.js#L117-L118

I have a fix proposal in the works now.

chrissrogers:
Good spec. We'll have the fix in soon.

bhelx:
@seebees Here is the code side of the fix: https://github.com/recurly/recurly-js/pull/199

Thanks for reporting!

PR #199, opened 2015-03-30
Fixes issue with tax rounding

chrissrogers:
##### Summary

- Instead of using a rounding method prone to floating point math issues, this uses a coercion using exp notation
- Tests provided in #198

### DIFF
--- a/lib/recurly/pricing/calculations.js
+++ b/lib/recurly/pricing/calculations.js
@@ -114,8 +114,8 @@ Calculations.prototype.tax = function (done) {
         });
 
         // tax estimation prefers partial cents to always round up
-        self.price.now.tax = Math.ceil(self.price.now.tax * 100) / 100;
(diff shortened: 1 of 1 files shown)


lib/recurly/pricing/calculations.js
```diff
@@ -237,3 +237,16 @@ Calculations.prototype.planPrice = function () {
 function decimal (prop) {
   this[prop] = (Math.round(Math.max(this[prop], 0) * 100) / 100).toFixed(2);
```
chrissrogers: I'm thinking I should fix this as well. It's less prone to side effects of floating point math problems because it's using a round and not a ceiling, but it is flawed nonetheless

chrissrogers: The method being used is ~36x faster, but the scientific coercion method is absolutely still within bounds of acceptable performance.

http://jsperf.com/decimal-rounding-0

bhelx:
:+1:

PR #202, opened 2015-05-05
Add support for ach payments with bank accounts

snodgrass23:
- adds bankAccount class 
    - includes `token` function for getting tokens with bank account info for ach payments
    - includes `bankInfo` function for getting bank info (name) using a routing number

### DIFF
--- a/lib/recurly/bank-account.js
+++ b/lib/recurly/bank-account.js
@@ -1,0 +1,179 @@
+
+/*!
+ * Module dependencies.
+ */
+
+var bind = require('bind');
+var each = require('each');
+var type = require('type');
+var index = require('indexof');
+var debug = require('debug')('recurly:bankAccount');
+var dom = require('../util/dom');
+var errors = require('../errors');
+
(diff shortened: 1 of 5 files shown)


lib/recurly/bank-account.js
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * expose
+ */
+
+module.exports = function bankAccount () {
+  return {
+    token: bind(this, token),
+    bankInfo: bind(this, bankInfo)
+  };
+};
```
chrissrogers: I'm not a huge fan of `Recurly.bankAccount().token` as the calling pattern. It would be more efficient to set up the binding once instead of on every call, and `bankAccount` should really be a module-like namespace instead of a callable function.

Can we export the unbound functions here, and then bind them in the main `Recurly` initializer?

chrissrogers:
Lookin' good. Let's get bank accounts in! :hamburger: :palm_tree:

Issue #167, opened 2014-12-18
Single use coupons are included in `discount_next` calculation


whiteotter:
when really they should only be included in the discount_now calculation

snodgrass23:
For clarification, the `discountNext` calculation assumes all coupons are equal and doesn't check to see if it is a one time coupon and shouldn't be applied to the estimate for the next invoice. This doesn't affect the actual coupon's use since the actual application is done on the Recurly servers and not in this library, which is only serving to provide an estimate.

snodgrass23:
The fix had been merged into the master branch and will be present in the next release.

PR #204, opened 2015-05-07
update discount calculations to respect single_use coupons

snodgrass23:
https://github.com/recurly/recurly-js/issues/167

When calculating discount for coupons, check single_use property before applying discount to next invoice estimate.


whiteotter:
+1 Tested Locally

PR #205, opened 2015-05-09
Adds prefix to jsonp callbacks

chrissrogers:
##### Summary

- Prevents collision with libraries using the default prefix
- See duojs/test#69
- Includes #207


bhelx:
:+1:

PR #207, opened 2015-05-11
Fixes an issue where XDRs aborted unpredictably

chrissrogers:
- XDRs are known to abort if too many are sent simultaneously
- This allows sends to execute asynchronously from their setup and space apart more.
- This will fix intermittent IE9 test failures
- See note on https://developer.mozilla.org/en-US/docs/Web/API/XDomainRequest


bhelx:
:+1:

PR #208, opened 2015-05-12
Cleans up token, bankAccount, and pricing modules

chrissrogers:
##### Summary

- To make [code climate happy](https://codeclimate.com/github/recurly/recurly-js/lib/recurly/token.js)
- Extracts normalizer into separate module
- Removes unnecessary routines
- Cleans up some syntax


sch:
If travis is happy, I’m happy

PR #210, opened 2015-05-14
Adds .jshintrc and corrects for JSHint

chrissrogers:
##### Summary

- Part of an effort to make code climate more useful
- Follow-up to #208
- This also adds a return to `dom.data` on both get and set calling patterns. This is simply a util API change and has no impact on the library functionality as a whole. I would still like to assess separately whether to adopt an external dom library or to pull our dom utils out into a separate lib.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,13 @@
+{
+  "curly": false,
+  "browser": true,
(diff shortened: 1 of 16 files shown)


sch:
:^o

PR #212, opened 2015-06-09
Additional currencies

csmb:
Adding new currencies Brazilian Real, Israeli New Sheqel, Indian Rupee, and Mexican Peso while also updating the version to 2.2.11

cc: @chrissrogers


chrissrogers:
:moneybag: :moneybag: :moneybag:

PR #213, opened 2015-06-11
Version 3.1.1

snodgrass23:
Updates for version 3.1.1 release.

cc: @chrissrogers


chrissrogers:
Just a heads up that we'll leave it open until next Wednesday, in case anything else comes in between now and then.

PR #214, opened 2015-06-16
Add trial calculations to discount next

rmclain:
If plan has trial adjust the discount to reflect the next payment for single use coupon codes.  Discount next was not getting updated to reflect the discount to be applied if the plan had a trial period.


rmclain:
After speaking with a rep from Recurly, it seems that a single use coupon is used up during a trail period.  So this logic isn't necessary.  Please ignore.

PR #221, opened 2015-08-27
Fix tests

cbarton:
- mocha @ master is currently not working, so let's use the most recent release. 
- the new test runner uses recurly.js not recurly.min.js, so let's build all files


snodgrass23:
:+1:

PR #231, opened 2016-02-05
Check if cvv is required before ignoring empty value

snodgrass23:
When validating input values, check if cvv is present in required fields before skipping an empty value.

### DIFF
--- a/test/validate.test.js
+++ b/test/validate.test.js
@@ -65,7 +65,7 @@ describe('Recurly.validate', function () {
     });
 
     it('should return true for leading zeros', function() {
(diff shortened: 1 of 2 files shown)


test/validate.test.js
```diff
@@ -65,7 +65,7 @@ describe('Recurly.validate', function () {
     });
 
     it('should return true for leading zeros', function() {
-      assert(true === recurly.validate.expiry('01', '16'));
+      assert(true === recurly.validate.expiry('01', '19'));
```
chrissrogers: ha!

chrissrogers:
+1

PR #243, opened 2016-03-30
Adds field integrity check to configure call

### LINKED ISSUE #239: Add DOM state check to configure call
`configure` should check the state of the stored fields and reset if the stored fields are no longer present.

Relevant [Stack Overflow question](https://stackoverflow.com/questions/35779175/issues-with-recurly-js-form-generation-and-gwt-ajax)

chrissrogers:
- if hosted field integrity is compromised, we reset
- Fixes #239


chrissrogers:
I'm closing this as a better solution lies in #267

PR #254, opened 2016-05-06
Adding data-collector module for fraud checking

snodgrass23:
Adds a fraud module that allows enabling Kount data collector. Also allows merchant to pass in a litle session id that will be used for vantiv threat matrix fraud checking on the transaction.

**Usage:**
Add the needed option to the config call:
``` js
recurly. configure({
    publicKey: 'PUBLIC_KEY',   
    fraud: {
       kount: { dataCollector: true },
       litle: { sessionId: 'my_session_id' }
    }
})
```

Also see: https://github.com/recurly/recurly-js-examples/pull/24 for an example of the format

cc: @karentobo


lib/recurly/fraud.js
```diff
@@ -0,0 +1,83 @@
...
+const debug = require('debug')('recurly:fraud');
+
+export class Fraud {
+  constructor (recurly) {
+    this.recurly = recurly;
+    this.selector = this.recurly.config.fields.number;
+    this.dataCollectorInitiated = false;
+    this.form = null;
+
+    recurly.ready(bind(this, this.dataCollector));
```
chrissrogers: We can use `Function.prototype.bind` now since we're IE9+

lib/recurly.js
```diff
@@ -35,6 +36,10 @@ const defaults = {
   publicKey: '',
   parent: true,
   cors: false,
+  fraud: {
+    dataCollector: false,
```
chrissrogers: It would be nice to expose less of the internals of fraud detection in the configuration api. Is there a name for this that would be more meaningful for a novice implementer? Perhaps something that directly matches any fraud configuration the merchant does in the app.

snodgrass23: I think that name is pretty meaningful. It's called data collector.  Here is the page from the Kount docs that all of our merchants would have went through when setting up their account with Kount.

It doesn't map to any configuration in the app because those are separate things. The app does it's piece independent of this and works the same whether this is in effect or not. This is simply an additional feature we're allowing them to enable that will give them more info on the Kount side.

screen shot 2016-05-11 at 9 36 51 am

lib/recurly/fraud.js
snodgrass23: Understandable. However, every other place in RJS is using the bind component so it seems for consistency at least we would use it here. I would think that we would either remove that component altogether now, or use it's inherent abstraction to update the details of the binding process inside of it.

chrissrogers: I've been churning it out -- new code doens't use it :)

lib/recurly/token.js
```diff
@@ -65,6 +66,8 @@ function token (options, done) {
   var data = normalize(fields, options, { parseCard: true });
   var inputs = data.values;
 
+  inputs.fraud = this.fraud.params(inputs);
+
```
chrissrogers: LEt's move this into the parent only block, since fraud data collection should only occur in that context.

test/fraud.test.js
```diff
@@ -0,0 +1,145 @@
...
+
+      assert(window.document.getElementById(testId) === null);
+
+      initializeRecurlyWith('successfulResponse');
+
+      assert(recurly.request.calledOnce);
+      assert(window.document.getElementById(testId) != null);
+    });
+  });
+});
```
chrissrogers: Radddd tests

test/token.test.js
```diff
@@ -208,6 +208,29 @@ apiTest(requestMethod => {
...
+            fraud: {
+              dataCollector: true,
+              litleSessionId: 'a0sd8f70a8s7df'
+            }
+          });
+          this.recurly.fraud.dataCollectorInitiated = true;
+          this.recurly.ready(done);
+        });
+
+        it('yeilds a token', function (done) {
```
chrissrogers: nit: yields

lib/recurly.js
chrissrogers: I can see the value in calling it dataColector within the Kount context, then. I'm a bit concerned with moving that into our broad 'fraud' context though. How about this as a compromise?

```js
{
  fraud: {
    kount: { dataCollector: false },
    litle: { sessionId: null }
  }
}
```

That way we're a bit more clear within the space of 'Recurly fraud protection'

lib/recurly.js
```diff
@@ -66,6 +71,7 @@ export class Recurly extends Emitter {
       token: bind(this, bankAccount.token),
       bankInfo: bind(this, bankAccount.bankInfo)
     };
+    this.fraud = new Fraud(this);
```
chrissrogers: I'm thinking this should just run in the parent context, otherwise it will execute within the hosted fields individually as well.

I'm thinking a refactor is going to help with the separation of concerns between parent and field contexts quite a bit. That should follow this.

lib/recurly/fraud.js
snodgrass23: sounds good!

lib/recurly.js
snodgrass23: It seems a bit overly nested since `kount` and `litle` are very likely to never have another option. It also forces usage of these config options to check for another level of presence, or at least the base Recurly class needs to always make sure the structure is there. I'm fine with it though if it makes it easier to understand.

lib/recurly.js
snodgrass23: Yeah, I had a bit of trouble deciphering where the parent stuff was, etc. I almost put it in the `parent()` function here:

https://github.com/recurly/recurly-js/blob/9b7fcf68f538d8e012f0ed020ca7bf7d9793c20d/lib/recurly.js#L155-L179

But, since everything else in that function is hosted fields related, I wasn't sure if it was a good place.

lib/recurly.js
chrissrogers: Should be okay if they're in the default here -- it'll deep merge in and this nesting level should be trusted to always be present.

lib/recurly.js
chrissrogers: Yeah totally. I think a concept of parent/child context will clarify that considerably.

test/token.test.js
```diff
@@ -246,7 +269,7 @@ apiTest(requestMethod => {
         });
 
         describe('when cvv is valid', function () {
-          it('produces a validation error', function (done) {
+          it('yields a token', function (done) {
```
chrissrogers: :]

chrissrogers:
There are currently test failures around injection. The code looks great to me, so I'll merge it once it goes green 💯

snodgrass23:
hmm.. tests were passing locally, but I was only running the new fraud tests. The couple of them fail when I run the whole suite. Must be some kind of scope leak or something going on. I'll check it out.

snodgrass23:
ahh.. Looks like my `fixture('minimal');` call got moved outside of the `beforeEach()` block when I was refactoring some of those tests and it didn't like it.

karentobo:
👍

snodgrass23:
We have a couple of random CI tests failing. Rerunning to try and get it green.

snodgrass23:
The tests pass a little more often when I remove the new fraud suite. Maybe something in the setup is effecting later tests because the suite itself is always passing. Commenting out the suite for now to get this in for a merchant that's waiting for it. Will post a followup PR to work through those tests and get them working in CI.

PR #257, opened 2016-05-12
Reconnects bus after reset (fixes field integrity check)

ewlloyd:
The bus.reset() call was removing the event listener -- this patch just reconnects it.

Please note my environment isn't set up to run the test suite on this, but it works well for my use case.

### DIFF
--- a/lib/recurly/bus.js
+++ b/lib/recurly/bus.js
@@ -84,6 +84,6 @@ export class Bus extends Emitter {
   reset () {
     this.recipients = [];
     window.removeEventListener('message', this.receive, false);
-    connect ();
+    this.connect ();
   }
 }
--- a/lib/recurly/bus.js
+++ b/lib/recurly/bus.js
@@ -77,12 +77,13 @@ export class Bus extends Emitter {
   /**
    * Resets the bus
    *  - Empties recipients
-   *  - disconnects listener
+   *  - cycles listener
    *
    * @public
    */
   reset () {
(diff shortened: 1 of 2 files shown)


ewlloyd:
Is your Travis setup pooched, or did I break something?

chrissrogers:
Great catch, thanks so much for this contrib. If you move the connect call into the bus `reset` method, I'll merge it right in. That will correct the method to actually reset, instead of 'disconnect'

ewlloyd:
Ubet. Wasn't sure if `bus.reset()` was in use elsewhere -- it doesn't appear to be.

I've also changed the doc comment to reflect the new behavior.

lib/recurly/bus.js
```diff
@@ -77,12 +77,13 @@ export class Bus extends Emitter {
...
    *  - Empties recipients
-   *  - disconnects listener
+   *  - cycles listener
    *
    * @public
    */
   reset () {
     this.recipients = [];
     window.removeEventListener('message', this.receive, false);
+    connect ();
```
chrissrogers: will need to be `this.connect();`

ewlloyd: Oh, the shame of it all...

Must've knocked it off while reviewing the diff. Sorry!

chrissrogers: No problem! I do it all the time :)

chrissrogers:
👍

PR #264, opened 2016-06-15
Fixes bug around initial pricing not respecting coupons

chrissrogers:
- If a coupon code was pre-populated into a priced field, it was not initially added to the pricing instance structure
- Removes `pricing.binding` alias for `pricing.attach`. It had been deprecated since Recurly.js v3 alpha phases
- Adds tests around pricing attachment

##### Repro steps

- Create a Recurly.js form with pricing and coupon input via form attachment
- Pre-populate a coupon field with a valid coupon code
- The pricing instance should take the coupon code into effect from the initial attachment


lib/recurly/pricing/attach.js
```diff
@@ -8,31 +8,23 @@ var find = require('component-find');
...
 
-exports.attach = attach;
-
-/**
- * Backward-compatibility
- *
- * @deprecated
- */
-
-exports.binding = attach;
```
chrissrogers: It was about time this went away

lib/recurly/pricing/attach.js
```diff
@@ -8,31 +8,23 @@ var find = require('component-find');
...
+export function attach (el) {
   var self = this;
   var elems = {};
 
   el = dom.element(el);
 
   if (!el) throw new Error('invalid dom element');
 
-  if (this.detatch) this.detatch();
+  if (this.detach) this.detach();
```
chrissrogers: That'd be a typo, Will Robinson

lib/recurly/pricing/attach.js
```diff
@@ -47,9 +39,9 @@ function attach (el) {
     events.bind(elem, 'propertychange', change);
   });
 
-  this.detatch = detatch;
+  this.detach = detach;
 
-  change();
+  change(INIT_RUN);
```
chrissrogers: The gist of this bug is that this initial `change` run wasn't triggering any changes because of the IE event target finder fallback. It should target the `<script>` element that is running this code as the target element, thus never triggering the initial run detection in the `target` method. This makes the intent far clearer IMO.

jayseeg:
+1

PR #267, opened 2016-06-16
Multiple instances

chrissrogers:
- Allows a the recurly singleton to spawn new instances without colliding with each others messages
- Allows an instance to be safely destroyed, removing its listeners and unbinding it from the DOM.


jayseeg:
+1

irnnr:
So how would one use this now?

chrissrogers:
The module's export is still a singleton `recurly`, but the internals allow for you to call the `recurly.configure` method many times with different targets for the hosted fields, and now the library will reinstantiate using the new targets.

irnnr:
Ok, I think that much was clear. But how are the different targets defined?  
Could the documentation be updated on that topic?

chrissrogers:
@irnnr simply call `recurly.configure` with different values for the field selectors.

```js
recurly.configure({
  // ...
  fields: {
    number: { selector: '#recurly-number-1' }
  }
});

recurly.configure({
  // ...
  fields: {
    number: { selector: '#recurly-number-2' }
  }
});
```

PR #268, opened 2016-06-16
MT-1574 - Remove UsageAddons from Calculations. Attach UsageAddon dat…

SteveDugas:
…a to pricing.base.usageAddons.


lib/recurly/pricing/calculations.js
```diff
@@ -176,6 +184,37 @@ Calculations.prototype.addons = function () {
...
+ *
+ * @private
+ */
+
+Calculations.prototype.usageAddons = function(){
+  var usageAddons = filter(this.items.plan.addons,function(addon){
+    return addon.add_on_type === "usage";
+  });
+
+  each(usageAddons, function (addon) {
```
chrissrogers: we can use `usageAddons.forEach` here. These shims were necessary when this module was first written, but out browser support doesn't necessitate it anymore.

lib/recurly/pricing/calculations.js
```diff
@@ -176,6 +184,37 @@ Calculations.prototype.addons = function () {
...
 /**
+ * Attaches usage addon info to the price object
+ *
+ * @private
+ */
+
+Calculations.prototype.usageAddons = function(){
+  var usageAddons = filter(this.items.plan.addons,function(addon){
+    return addon.add_on_type === "usage";
+  });
```
chrissrogers: Same note here on `Array.prototype.filter` -- let's use the native method.

lib/recurly/pricing/calculations.js
```diff
@@ -176,6 +184,37 @@ Calculations.prototype.addons = function () {
...
+      price = addon.price[this.items.currency].unit_amount;
+    } else {
+      percentage = addon.usage_percentage;
+    }
+
+    this.price.base.usageAddons[addon.code] = {
+      price: price,
+      measured_unit: name,
+      percentage: percentage
+    };
```
chrissrogers: style nit

```js
{
  price,
  percentage,
  measured_unit: name
}
```

lib/recurly/pricing/calculations.js
```diff
@@ -176,6 +184,37 @@ Calculations.prototype.addons = function () {
...
+Calculations.prototype.usageAddons = function(){
+  var usageAddons = filter(this.items.plan.addons,function(addon){
+    return addon.add_on_type === "usage";
+  });
+
+  each(usageAddons, function (addon) {
+    var usageType = addon.usage_type;
+    var name = addon.name;
+    var price = null;
+    var percentage = null;
```
chrissrogers: nit: use `const` and `let`

chrissrogers:
Noting the fixture update is good. If we add a few spec updates to go with it, this will be good to go.

SteveDugas:
@chrissrogers Updated, now with a few tests.

PR #269, opened 2016-06-23
Enables CORS by default

chrissrogers:
- 🍺 
- Also removes a `setTimeout` from the fraud tests, ~20% test speed increase


csmb:
👀👍🏼

PR #272, opened 2016-07-06
Don't try to calculate addons if there aren't any addons defined for …

SteveDugas:
…the plan.

### DIFF
--- a/lib/recurly/pricing/calculations.js
+++ b/lib/recurly/pricing/calculations.js
@@ -160,6 +160,10 @@ Calculations.prototype.addons = function () {
   this.price.now.addons = 0;
   this.price.next.addons = 0;
   // exclude usage addons.
+  if (!Array.isArray(this.items.plan.addons)) {
+    return;
+  }
+
   let fixedAddons = this.items.plan.addons.filter(function(addon){
     return addon.add_on_type === "fixed";
   });


lib/recurly/pricing/calculations.js
```diff
@@ -69,9 +69,11 @@ Calculations.prototype.subtotal = function () {
...
   this.price.next.subtotal += this.price.next.plan;
 
-  this.addons();
-  this.price.now.subtotal += this.price.now.addons;
-  this.price.next.subtotal += this.price.next.addons;
+  if(this.items.plan.addons && this.items.plan.addons.length){
+    this.addons();
+    this.price.now.subtotal += this.price.now.addons;
+    this.price.next.subtotal += this.price.next.addons;
+  }
```
chrissrogers: This needs to move down into the addons method, since it initially sets the addons values to zero. I recommend bailing after that initial set if there are no addons.

lib/recurly/pricing/calculations.js
```diff
@@ -160,6 +160,10 @@ Calculations.prototype.addons = function () {
   this.price.now.addons = 0;
   this.price.next.addons = 0;
   // exclude usage addons.
+  if( !Array.isArray(this.items.plan.addons) ){
```
chrissrogers: nit: `if (expr) {`, just to match the code style of the rest of the module.

chrissrogers:
+1

PR #276, opened 2016-07-21
Parallelizes travis tests

chrissrogers:
- 💇 
- runs slightly faster
- better test isolation, so that a failure in one browser env doesn't require a rerun of all
- artifacts now require passage of phantom tests and not all


csmb:
+1

PR #279, opened 2016-07-22
Allows a label tag paired to a field container to defer focus to the field

### LINKED ISSUE #240: Is it possible to programmatically focus field in iframe?
Currently also `<label>` doesn't cause the field to focus. Is there way how to enable that? It's nice from user perspective :).

### LINKED ISSUE #271: Focus input fields on click, tap, and focus within field containers
- From #260

chrissrogers:
- Finds labels paired to containers via for-id match
- Sends `hostedField:${type}:focus!` event when label is clicked/tapped
- This will not work with mobile Safari due to strict focus event constraints. See [this][so]
- Fixes #240
- Fixes #271

[so]: http://stackoverflow.com/questions/12204571/mobile-safari-javascript-focus-method-on-inputfield-only-works-with-click#answer-16601288

### DIFF
--- a/lib/recurly/hosted-field.js
+++ b/lib/recurly/hosted-field.js
@@ -1,4 +1,5 @@
 import Emitter from 'component-emitter';
+import events from 'component-event';
 import dom from '../util/dom';
 import errors from '../errors';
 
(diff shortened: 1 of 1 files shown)


bhelx:
:+1:

PR #282, opened 2016-07-27
Adds safety check to dom value get method

### LINKED ISSUE #220: Add safety check to valueGet function
https://github.com/recurly/recurly-js/blob/master/lib/util/dom.js#L82

Need a safety check here to prevent an error on page load:
```value = node.options[node.selectedIndex].value;```

Without it there is the possibility of getting an error like this on page load:

Firefox:
```TypeError: node.options[node.selectedIndex] is undefined recurly.js:1:25832```
Chrome:
```Uncaught TypeError: Cannot read property 'value' of undefined recurly.js:1```
IE:
```Unable to get property 'value' of undefined or null reference File: recurly.js, Line: 1, Column: 25833```

### PR DESCRIPTION
**Pull request by chrissrogers**
- If a merchant uses a select input with no options in a pricing attachment (see example below), the `valueGet` function would blow up as described in #220.
- Fixes #220

```html
<form>
  <select data-recurly="plan">
  </select>
</form>
```

```js
let pricing = recurly.Pricing();
pricing.attach(document.querySelector('form'));

// ... retrieve plans asynchronously and add options to the plan select input
```

### DIFF
--- a/lib/util/dom.js
+++ b/lib/util/dom.js
@@ -80,7 +80,8 @@ function valueGet (node) {
   if (!nodeType) {
     value = '';
   } else if ('options' in node) {
-    value = node.options[node.selectedIndex].value;
+    let option = node.options[node.selectedIndex] || node.options[0] || { value: '' };
+    value = option.value;
   } else if (nodeType === 'checkbox') {
     if (node.checked) value = node.value;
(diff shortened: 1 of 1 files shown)


jayseeg:
👍🏻

PR #288, opened 2016-08-02
Fixes bug with pricing attachment of addons

chrissrogers:
- Addons were being set before plans might be available on pricing.
- Attachment assumes a `PricingPromise` at all times during update. This ensures it has one.


SteveDugas:
👍

PR #293, opened 2016-08-22
Add support for tabbing in a Mobile browser.

SteveDugas:
This PR does 2 things to add mobile tabbing between hosted fields support.

1) Adds a "hidden" input before the iFrame. iOS Safari at least doesn't tab into the iframe like most desktop browsers do. The hidden input just transfers the focus from itself to the hosted field inside.

2) Adds support for hostedField:tab:previous and hostedField:tab:next events. These events just grab all [data-recurly] elements (either actual input fields or a div containing a hosted field), and then transfer focus to whichever one is previous or next.

### DIFF
--- a/lib/recurly/hosted-field.js
+++ b/lib/recurly/hosted-field.js
@@ -28,6 +28,7 @@ export class HostedField extends Emitter {
     this.configure(options);
     this.inject();
     this.bindLabel();
+    this.bindTabTrap();
 
     this.on('bus:added', bus => {
       this.bus = bus;
@@ -69,11 +70,14 @@ export class HostedField extends Emitter {
 
   inject () {
     this.target.innerHTML = `
-      <div class="${this.classList()}">
+      <div class="${this.classList()}" style="position:relative;">
+        <input type="text" style="position: absolute; z-index: -1; width: 0; height: 0;" data-recurly-hosted-field="true" />
         <iframe
           src="${this.url()}"
+          data-recurly-iframe="true"
           border="0"
           frameborder="0"
+          tabIndex="-1"
           allowtransparency="true"
           scrolling="no">
         </iframe>
@@ -81,12 +85,15 @@ export class HostedField extends Emitter {
     `;
 
     this.container = this.target.children[0];
-    this.iframe = this.container.children[0];
+    this.tabTrapInput = this.container.children[0];
+    this.iframe = this.container.children[1];
     this.window = this.iframe.contentWindow;
 
     this.iframe.style.height = '100%';
     this.iframe.style.width = '100%';
     this.iframe.style.background = 'transparent';
+    this.iframe.style.position = 'relative';
+    this.iframe.style.zIndex = 1;
   }
 
   bindLabel () {
@@ -100,9 +107,22 @@ export class HostedField extends Emitter {
   reset () {
     this.off();
     this.target.innerHTML = '';
+    // TODO: May need to remove tabTrap stuff
     delete this.target;
     delete this.iframe;
     delete this.window;
+  }
+
+  bindTabTrap () {
+    this.tabTrapInput.addEventListener('focus',e => {
+      e.preventDefault()
+      e.stopPropagation()
+      // Firefox demands a blur or else the focus doesn' work...
+      if(navigator.userAgent.toLowerCase().indexOf('firefox') >= 0){
+        e.currentTarget.blur()
+      }
+      this.focus()
+    })
   }
 
   update () {
--- a/lib/recurly/hosted-fields.js
+++ b/lib/recurly/hosted-fields.js
@@ -2,8 +2,11 @@ import clone from 'component-clone';
 import Emitter from 'component-emitter';
 import merge from 'lodash.merge';
 import omit from 'lodash.omit';
+import indexOf from 'lodash.indexof'
 import {HostedField} from './hosted-field';
 import errors from '../errors';
+
+const FOCUS_SELECTOR = 'a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), iframe:not([data-recurly-iframe]), object, embed, [tabindex="0"], [contenteditable]';
 
 const debug = require('debug')('recurly:hostedFields');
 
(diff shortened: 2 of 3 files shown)


lib/recurly/hosted-field.js
```diff
@@ -82,6 +87,16 @@ export class HostedField extends Emitter {
     });
   }
 
+  bindTabTrap () {
+    this.tabTrapInput.addEventListener('focus',(e) => {
+      e.preventDefault()
+      e.stopPropagation()
+      // Firefox demands a blur or else the focus doesn' work...
+      if(typeof InstallTrigger !== 'undefined') e.currentTarget.blur()
```
chrissrogers: Is the `InstallTrigger` detection a generic FF detection, or somehow related to the need to blur before refocusing?

chrissrogers:
nit: add semicolons and match statement spacing patterns. Can always go back and fix that tho

lib/recurly/hosted-field.js
```diff
@@ -82,6 +87,16 @@ export class HostedField extends Emitter {
     });
   }
 
+  bindTabTrap () {
+    this.tabTrapInput.addEventListener('focus',(e) => {
```
chrissrogers: nit: no need for the parens around `e`

lib/recurly/hosted-fields.js
```diff
@@ -1,6 +1,7 @@
 import omit from 'lodash.omit';
 import merge from 'lodash.merge';
 import Emitter from 'component-emitter';
+import { indexOf } from 'lodash'
```
chrissrogers: nit: spacing

lib/recurly/hosted-fields.js
```diff
@@ -29,13 +30,55 @@ export class HostedFields extends Emitter {
...
     this.configure(options);
     this.inject();
     this.on('hostedField:state:change', this.update.bind(this));
     this.on('bus:added', bus => {
       this.bus = bus;
       this.fields.forEach(hf => bus.add(hf));
     });
+
+    this.on('hostedField:tab:previous',this.onTabPrevious)
+    this.on('hostedField:tab:next',this.onTabNext)
```
chrissrogers: Since these handlers are only called on these events, let's bind them here.

```js
this.on('hostedField:tab:previous',this.onTabPrevious.bind(this);
```

lib/recurly/hosted-fields.js
```diff
@@ -29,13 +30,55 @@ export class HostedFields extends Emitter {
...
+  }
+
+  onTabNext(body) {
+    let type = body.type
+    let field = this.fields.find(function(f){ return f.type === type })
```
chrissrogers: Might as well use an arrow fn here, and a const since it's not meant to change.

```js
const field = this.fields.find(f => f.type === type);
```

lib/recurly/hosted-fields.js
```diff
@@ -29,13 +30,55 @@ export class HostedFields extends Emitter {
...
+    let allFields = document.querySelectorAll("[data-recurly]")
```
chrissrogers: So the one issue this may present is if the fields are not bound using this selector. It's possible for the merchant to specify other selectors. I think we'll need to incorporate the fields known by the `HostedFields` instance.

lib/recurly/hosted-field.js
```diff
@@ -54,7 +55,8 @@ export class HostedField extends Emitter {
 
   inject () {
     this.target.innerHTML = `
-      <div class="${this.classList()}">
+      <div class="${this.classList()}" style="position:relative;">
+        <input type="text" style="position: absolute; z-index: -1; width: 0; height: 0;" />
```
SteveDugas: Yup, takes it out of tab order.

lib/recurly/hosted-field.js
SteveDugas: generic FF detection for FF 1.0+. It's the method they use for installing Addons or something.

lib/recurly/hosted-field.js
chrissrogers: Ohh css

lib/recurly/hosted-field.js
chrissrogers: I'm a little skeptical of the single deduction. Could we do userAgent detection just as easily, with the intent being a bit more clear?

Something like `~navigator.userAgent.toLowerCase().indexOf('firefox')`

lib/recurly/hosted-fields.js
chrissrogers: It doesn't, but it would be able to tell you the html elements of the hosted fields in that instance. hostedFields.fields contains the array of HostedField instances, which each ref their containers. You could then crawl the dom tree for the previous and next tabbing targets in the form

SteveDugas: What do you think the best way to do that is? The previous and next ones that aren't hosted fields are pretty difficult to find, or at least I don't see an easy way to figure that out. If the elements have a taborder attribute it's not too difficult, but I'm not sure exactly how that would work in practice. Would clients put taborder on like the parent div of hosted fields also?  I think that would probably mess up the way I'm using a fake input before the first hosted field (outside the iframe) to have something to tab into.

SteveDugas: Outside of using an attribute like tabindex I'm not sure the best way to find what the next or previous tabable element should be. I'd need to reimplement the rules for which elements are tabbable to get the right set, and then what order the iframes fit in. I started down that road a little but it seemed a bit ridiculous for me. I thought it would be better to have a way for support to tell people how to make sure tabbing works (hey, make sure you have [recurly-data] on your recurly hosted field divs and at least the immediate elements before and after.)

package.json
```diff
@@ -67,6 +67,7 @@
     "koa-jsonp": "^0.1.2",
     "koa-logger": "^1.3.0",
     "koa-route": "^2.4.2",
+    "lodash": "^4.15.0",
```
chrissrogers: Let's use lodash.indexof to keep the package size as small as possible.

lib/recurly/hosted-fields.js
chrissrogers: It's entirely possible for a hosted field to be the last element in the form, thus no [data-recurly] elements would follow. I think we can crawl the DOM looking for the next/prev element that is focusable (or a hosted field) and push focus to that element. jquery-ui has a `:focusable` selector method [here](https://github.com/jquery/jquery-ui/blob/89af4c292eaa5fc1c83437ca71085264dcbef34a/ui/focusable.js#L28) that implements the spec's definition of what can be focused. We can drop an analogous method into the rjs dom lib that also looks for hosted fields (which the `recurly.hostedFields` instance would tell us). It's certainly more complex, but I think it would be the best solution to make the feature work everywhere

chrissrogers: Actually [this](https://github.com/jkup/focusable) is pretty clever. It just lists the selector queries to find the focusable elements. We can use that with the recurly field selectors given in `recurly.config.fields[field].selector` to find the next focusable target.

SteveDugas: I changed this to basically use what the output of focusable was giving me, but made a change so it didn't include our iframes (but does include my fake tab trapping input).

SteveDugas:
Updated this to reflect your review. The hosted-fields one is still the same.

chrissrogers:
+1

PR #295, opened 2016-08-30
Sends initial field config over hash instead of query

chrissrogers:
- This will let us cache those responses more aggressively


snodgrass23:
👍

drewish:
Do we need to version this? It seems like we'd need some overlap between styles during the rollout.

chrissrogers:
@drewish yes versioning is always required and is committed to master once the new version is prepared.

PR #302, opened 2016-10-18
Fixes issue with address attachment

chrissrogers:
- When a form has only a billing address, fixes an issue where shipping address would be set as blank and override billing taxes


jayseeg:
+1

PR #303, opened 2016-10-31
Ensures configuration objects are not externally mutable

chrissrogers:
- Fixes issue where external configuration assignments could undermine integrity checks


jayseeg:
👍🏻

PR #307, opened 2017-01-03
Include the responseText when json is invalid

geoffdutton:
We've seen some errors where the Recurly API must have responded with invalid JSON, but we haven't been able to see it live in order to see what the responseText actually is, so it'd would be helpful to include that in the error message.


chrissrogers:
This is absolutely helpful, especially for those older browsers where inspecting the network responses is not straightforward. Thanks for the contrib!

chrissrogers:
Phantom tests passed, but all browser tests require auth with sauelabs that can only be performed from the recurly github org.

PR #320, opened 2017-02-02
Also send name, address, etc fields for apple pay token creation

snodgrass23:
When sending Apple Pay payment data for a token, include the name and address fields. These are required to later build a valid billing info so that the customer can be billed successfully.

cc: @chrissrogers

### DIFF
--- a/lib/recurly/apple-pay.js
+++ b/lib/recurly/apple-pay.js
@@ -1,7 +1,9 @@
 import Promise from 'promise';
 import Emitter from 'component-emitter';
 import errors from '../errors';
+import dom from '../util/dom';
 import {Pricing} from './pricing';
+import {normalize} from '../util/normalize';
 
 const debug = require('debug')('recurly:apple-pay');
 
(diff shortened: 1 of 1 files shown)


jss79:
+1

lib/recurly/apple-pay.js
```diff
@@ -19,6 +21,28 @@ export function factory (options) {
...
+  'address2',
+  'country',
+  'city',
+  'state',
+  'postal_code',
+  'phone',
+  'vat_number',
+  'fraud_session_id',
+  'token'
+];
```
chrissrogers: I think it would be best to import these from the token module

lib/recurly/apple-pay.js
```diff
@@ -238,7 +262,17 @@ class ApplePay extends Emitter {
   onPaymentAuthorized (event) {
     debug('Payment authorization received', event);
 
-    this.recurly.request('post', '/apple_pay/token', event.payment.token, (err, token) => {
+    // TODO: can we be sure an apple pay form will always have the number field?
+    // var selector = this.recurlyconfig.fields.number.selector;
+    var selector = '[data-recurly=first_name]'
+    var form = dom.findNodeInParents(window.document.querySelector(selector), 'form');
+    var data = normalize(fields, form);
+    var inputs = data.values || {};
```
chrissrogers: I think we should have the form element passed in to the ApplePay class on init in order to avoid this loose coupling of the `data-recurly` convention. `options.form` seems best.

lib/recurly/apple-pay.js
snodgrass23: Agreed. I just didn't see a simple way to do it and wanted to get something out asap for Karen to test.

lib/recurly/apple-pay.js
snodgrass23: yeap, I was thinking about this too. It also fell to getting something out for Karen to test while deciding how best to get this in (I left the TODO comment as I knew it would draw your attention and you would probably have something to say on the subject). I'll use your idea of passing it in the options to the ApplePay class.

PR #321, opened 2017-02-09
fix test that is now failing because expiry isn’t a future date anymore

snodgrass23:
Attempting to get the v3 build to complete.


chrissrogers:
To the future!

PR #322, opened 2017-02-09
Update Apple Pay init to receive payment form

snodgrass23:
This will allow RJS to also pull out info from address and name fields to add to token.


jss79:
+1

Issue #329, opened 2017-02-20
window.addEventListener('message') should verify the message origin


jchayan:
I notice you are using the postMessage API for cross-origin communication. In my application there are other programs implementing this as well, which causes these messages to enter in your event listener (`relay.js:28`: `events.bind(window, 'message', listener)`).

In my use case, I'm receiving strings via postMessage, which causes `relay.js:29`: `var data = JSON.parse(event.data)` to fail. There should exist origin checking in order to avoid conflicts, e.g:

```javascript
events.bind(window, 'message', function listener (event) {
    if (event.origin === 'https://api.recurly.com') { /* or whatever property that holds this value */
        var data = JSON.parse(event.data);
        var name = data.recurly_event;
        var body = data.recurly_message;
        var err = body.error ? errors('api-error', body.error) : null;
        events.unbind(window, 'message', listener);
        if (name) self.emit(name, err, body);
        if (frame) document.body.removeChild(frame);
    }
});
```

jchayan:
I can merge the fix for this

jchayan:
@chrissrogers This should be closed :)

PR #330, opened 2017-02-20
Patch for issue #329

jchayan:
Checks origin before accepting messages

### DIFF
--- a/lib/recurly.js
+++ b/lib/recurly.js
@@ -50,7 +50,6 @@ const defaults = {
     litle: { sessionId: null }
   },
   api: 'https://api.recurly.com/js/v1',
-  origin: 'https://api.recurly.com',
   fields: {
     all: {
       style: {}
@@ -133,7 +132,6 @@ export class Recurly extends Emitter {
 
     if (options.api) {
       this.config.api = options.api;
-      this.config.origin = (new URL(options.api)).origin;
     }
 
     if (options.currency) {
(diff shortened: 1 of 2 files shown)


lib/recurly/relay.js
```diff
@@ -26,13 +26,15 @@ function relay (done) {
...
 
   events.bind(window, 'message', function listener (event) {
-    var data = JSON.parse(event.data);
-    var name = data.recurly_event;
-    var body = data.recurly_message;
-    var err = body.error ? errors('api-error', body.error) : null;
-    events.unbind(window, 'message', listener);
-    if (name) self.emit(name, err, body);
-    if (frame) document.body.removeChild(frame);
+    if (event.origin === 'https://api.recurly.com') {
```
chrissrogers: An early return would convey better flow control here.

chrissrogers: Technically the intended origin can be any base domain provided by the `api` property passed to config. The property is available here as `this.config.api`, from which the base uri can be extracted.

jchayan: @chrissrogers Understood, sorry for the late reply. Will be updating this soon.

lib/recurly.js
```diff
@@ -132,6 +133,7 @@ export class Recurly extends Emitter {
 
     if (options.api) {
       this.config.api = options.api;
+      this.config.origin = (new URL(options.api)).origin;
```
chrissrogers: Since this isn't meant to be configured, let's just compute this when it's needed by the relay.

I believe the `URL` class is not supported by Internet Explorer, which would preclude us using it. We could use a polyfill or add a util function that uses the `createElement` method.

chrissrogers:
Perfect, thanks very much for contributing this and for taking feedback into consideration!

PR #338, opened 2017-03-09
Sends Braintree device fingerprint along with PayPal requests

chrissrogers:
- If a device fingerprint can be obtained, attach it to PayPal tokenization requests


adeitrick:
👍

PR #339, opened 2017-03-13
Makes origin checking compatible with IE11 + possibly others

davedevelopment:
The `origin` attribute isn't available in IE11 on Windows 7, I haven't checked other IE versions, but I suspect the same. 

If anyone has access to multiple versions, they can check with this. https://jsfiddle.net/oLtyn55q/6/

The patch should at least make things work for more browsers than the current implementation does.

### DIFF
--- a/lib/recurly/relay.js
+++ b/lib/recurly/relay.js
@@ -28,8 +28,17 @@ function relay (done) {
   self.origin = function () {
     var parser = document.createElement('a');
     parser.href = self.config.api;
-
-    return parser.origin;
+    
+    if (parser.origin) {
+      return parser.origin;
+    }
+    
+    // IE11 seems to append :443 to parser.host even when it's not there
+    if (parser.protocol === 'https:' && parser.port == 443) {
+      return parser.protocol+'//'+parser.hostname;
+    }
+    
+    return parser.protocol+'//'+parser.host;
   };
 
   events.bind(window, 'message', function listener (event) {


lib/recurly/relay.js
```diff
@@ -28,8 +28,17 @@ function relay (done) {
...
+    if (parser.origin) {
+      return parser.origin;
+    }
+    
+    // IE11 seems to append :443 to parser.host even when it's not there
+    if (parser.protocol === 'https:' && parser.port == 443) {
+      return parser.protocol+'//'+parser.hostname;
+    }
+    
+    return parser.protocol+'//'+parser.host;
```
chrissrogers: nitpick: better to use string interpolation here and on line 38: 

```js
return `${parser.protocol}//${parser.host}`;

Comment (chrissrogers) [inline lib/recurly/relay.js]: What's the reason we'd want to use the `host` property instead of `hostname`?

Comment (davedevelopment) [inline lib/recurly/relay.js]: `host` would include the port, so `localhost:3000`, which is what you would normally want, but for some reason, IE11 includes `:443` if the protocol is `https`. So 

``` javascript
var parser = document.createElement('a');
parser.href = 'https://api.recurly.com/v1/';
console.log(parser.host); // api.recurly.com:443
```

This wouldn't work for the comparison, however I'm about to push another PR that will use the same function on both urls, which should make it more robust.

davedevelopment:
Closing in favour of #341

PR #341, opened 2017-03-14
Makes origin checking compatible with IE11 + possibly others

davedevelopment:
Builds on #340 and supercedes #339


chrissrogers:
oh fantastic, thanks for rebasing atop #340

chrissrogers:
+1 this is great -- thanks for the contrib!

PR #345, opened 2017-03-28
Fixes issue with new PayPal API usage of deprecated method

chrissrogers:
- Callbacks were improperly registered with the deprecated PayPal invocation


bhelx:
:+1:

PR #363, opened 2017-07-06
Adds support for Braintree deviceData

chrissrogers:
- Allows a merchant to pass Braintree fraud data collection data into credit card tokenization requests.

### DIFF
--- a/lib/recurly/token.js
+++ b/lib/recurly/token.js
@@ -1,6 +1,7 @@
 import each from 'component-each';
 import type from 'component-type';
 import {normalize} from '../util/normalize';
+import {uuid} from '../util/uuid';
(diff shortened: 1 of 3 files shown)


lib/recurly/token.js
```diff
@@ -123,7 +125,7 @@ function validate (input) {
   }
 
   each(this.config.required, function(field) {
-    if (!input[field] && ~fields.indexOf(field)) {
+    if (!input[field] && ~FIELDS.indexOf(field)) {
```
snodgrass23: do you have access to the native `includes` here?

chrissrogers: Unfortunately it isn't supported by IE

PR #367, opened 2017-07-21
Adds a missing error type

chrissrogers:
- For when Braintree is misconfigured within Recurly [here](https://github.com/recurly/recurly-js/blob/master/lib/recurly/paypal/strategy/braintree.js#L100)

cc @karentobo


snodgrass23:
Do we need to add any new mappings for the updated Apple Pay errors?

chrissrogers:
@snodgrass23 Those should pass in straight away. This is just to actually have a registered error where the client assumed one was.

PR #370, opened 2017-08-10
Adds support for starting an http server

chrissrogers:
- When using a proxy server during local development, it will be terminating tls connections instead of the app server


chrissrogers:
@adeitrick ping

adeitrick:
+1

PR #380, opened 2017-10-07
use Apple Pay contact data when form name and address are not supplied

snodgrass23:
If the merchant doesn't supply a form for the user to specify name and address data, attempt to use data present in Apple Pay contact data.

**Testing notes**
- [ ] When form fields are present, they are given priority and Apple Pay address is ignored.
- [ ] When Apple Pay card is displayed in Safari, it now asks for the address to use with the payment
- [ ] For fields not specified in purchase form, the address data specified in the Apple Pay popup card is used and passed.

### DIFF
--- a/lib/recurly/apple-pay.js
+++ b/lib/recurly/apple-pay.js
@@ -180,9 +180,10 @@ class ApplePay extends Emitter {
     inputs.paymentMethod = data.token.paymentMethod;
 
     if (!data.billingContact) return;
+    if (FIELDS.some(field => inputs[field])) return;
 
     FIELDS.forEach(field => {
-      if (inputs[field] || !APPLE_PAY_ADDRESS_MAP[field]) return;
+      if (!APPLE_PAY_ADDRESS_MAP[field]) return;
 
       let tokenData = data.billingContact[APPLE_PAY_ADDRESS_MAP[field]];
 
--- a/lib/recurly/apple-pay.js
+++ b/lib/recurly/apple-pay.js
@@ -8,6 +8,16 @@ const debug = require('debug')('recurly:apple-pay');
 
 const SESSION_TIMEOUT = 300000; // 5m
 const APPLE_PAY_API_VERSION = 2;
+const APPLE_PAY_ADDRESS_MAP = {
+  first_name: 'givenName',
+  last_name: 'familyName',
+  address1: 'addressLines',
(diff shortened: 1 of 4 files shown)


lib/recurly/apple-pay.js
```diff
@@ -8,6 +8,16 @@ const debug = require('debug')('recurly:apple-pay');
 
 const SESSION_TIMEOUT = 300000; // 5m
 const APPLE_PAY_API_VERSION = 2;
+const BILLING_DATA_MAP = {
```
chrissrogers: let's rename this `APPLE_PAY_ADDRESS_MAP` or something similar to be more specific.

lib/recurly/apple-pay.js
```diff
@@ -158,6 +169,23 @@ class ApplePay extends Emitter {
...
+    inputs.paymentMethod = data.token.paymentMethod;
+
+    FIELDS.forEach(field => {
+      if (inputs[field] || !BILLING_DATA_MAP[field]) return;
+
+      let tokenData = data.billingContact[BILLING_DATA_MAP[field]];
+
+      // address lines are an array from Apple Pay
+      if (field === 'address1') tokenData = tokenData[0];
+      else if (field === 'address1') tokenData = tokenData[1];
```
chrissrogers: should this be `'address2'`?

lib/recurly/apple-pay.js
snodgrass23: I thought about something like that, but having Apple Pay in the name in the Apple Pay class seemed redundant.

lib/recurly/apple-pay.js
snodgrass23: fixed this typo and added some basic specs

lib/recurly/apple-pay.js
```diff
@@ -159,6 +170,32 @@ class ApplePay extends Emitter {
...
+   * object that is sent to RA for tokenization
+   *
+   * @public
+   */
+  mapPaymentData (inputs, data) {
+    inputs.paymentData = data.token.paymentData;
+    inputs.paymentMethod = data.token.paymentMethod;
+
+    if (!data.billingContact) return;
+    if (FIELDS.some(field => { return inputs[field]; })) return;
```
chrissrogers: nit: `field => inputs[field]`

PR #382, opened 2017-10-20
CheckoutPricing: Subscriptions & Adjustments

chrissrogers:
#### Description

- `Pricing` is now parent class to `SubscriptionPricing` (alias `recurly.Pricing.Subscription`)
- Adds `CheckoutPricing` at alias `recurly.Pricing.Checkout`
- Adds `CheckoutPricing#subscription` and `CheckoutPricing#adjustment`
- Adds CheckoutPricing Calculations class, and handlers for subscriptions and adjustment subtotaling
- Refactors `PricingPromise`
- Does not include `SubscriptionPricing` change event linkage to `CheckoutPricing`

#### Example

```js
var subscriptionPricing = recurly.Pricing.Subscription();
var checkoutPricing = recurly.Pricing.Checkout();

checkoutPricing.adjustment({
  amount: 100,
  quantity: 1,
  // code: 'my-arbitrary-identifier',
  // taxExempt: true,
  // taxCode: 'my-tax-code'
});

subscriptionPricing.plan('my-plan-code').done(function () {
  // Multiple SubscriptionPricings may be added to CheckoutPricing
  checkoutPricing.subscription(subscriptionPricing).done(function (price) {
    // Print price output object to console
    console.log(price);
  });
});

```


lib/util/decimalize-member.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Applies a decimal transform on an object's member
+ *
+ * @param {String} prop Property on {this} to transform
+ * @this {Object} on which to apply decimal transformation
+ */
+
+export default function decimalizeMember (prop) {
+  this[prop] = (Math.round(Math.max(this[prop], 0) * 100) / 100).toFixed(2);
```
snodgrass23: does this need to account for zero decimal currencies?

chrissrogers: We might consider that for another change... it would add '.00' to a ZDC, which isn't proper for display, but the underlying values are sound.

PR #385, opened 2017-11-02
CheckoutPricing: Coupons

chrissrogers:
#### Description

- **To follow #382** -- includes CheckoutPricing subscription and adjustment framework
- Adds `CheckoutPricing#coupon`
- Adds `CheckoutPricing/Calculations#coupon` calculation routines

#### Example

```js
var subscriptionPricing = recurly.Pricing.Subscription();
var checkoutPricing = recurly.Pricing.Checkout();

subscriptionPricing.plan('my-plan-code').done(function () {
  // One coupon may be used at a time
  checkoutPricing
    .subscription(subscriptionPricing)
    .coupon('my-coupon-code')
    .done(function (price) {
      console.log(price);
    });
});

```


lib/recurly/pricing/subscription/calculations.js
```diff
@@ -182,19 +181,23 @@ export default class Calculations {
...
+      var discountNext = parseFloat((this.price.next.subtotal * coupon.discount.rate).toFixed(6));
+      this.price.now.discount = Math.round(discountNow * 100) / 100;
+      this.price.next.discount = Math.round(discountNext * 100) / 100;
+    } else if (coupon.discount.type === 'free_trial') {
+      // Handled in separate trial logic
+    } else {
+      const discountableNow = this.price.now.subtotal + this.price.now.setup_fee;
+      const discountableNext = this.price.next.subtotal;
+      this.price.now.discount = Math.min(discountableNow, coupon.discount.amount[this.items.currency]);
+      this.price.next.discount = Math.min(discountableNext, coupon.discount.amount[this.items.currency]);
```
chrissrogers: This is an important change to current pricing behavior.

Previously, if a gift card with a value higher than the subscription total (now + next cycle) was applied, it would apply a discount larger than the total. Now, the discount amount is limited to the total per cycle.

test/pricing/subscription/subscription.test.js
```diff
@@ -346,7 +346,7 @@ describe('Recurly.Pricing.Subscription', function () {
         .coupon('coop')
         .done(function (price) {
           assert.equal(price.now.discount, '20.00');
-          assert.equal(price.next.discount, '20.00');
+          assert.equal(price.next.discount, '19.99');
```
chrissrogers: See https://github.com/recurly/recurly-js/pull/385/files#diff-b1d0034ab4f6df9a3774c0f906e44a90R197

lib/recurly/pricing/subscription/calculations.js
```diff
@@ -63,11 +63,10 @@ export default class Calculations {
     this.price.now.subtotal += this.price.now.addons;
     this.price.next.subtotal += this.price.next.addons;
 
+    this.setupFee();
```
chrissrogers: This was moved prior to discount calculation in order to determine whether a coupon discount amount has exceeded the discountable subtotal for the subscription cycles.

lib/recurly/apple-pay.js
```diff
@@ -1,12 +1,12 @@
 import Emitter from 'component-emitter';
 import errors from '../errors';
+import {CheckoutPricing} from './pricing/checkout';
```
snodgrass23: Is this going to be used in one of the followup PR's?

lib/recurly.js
```diff
@@ -279,6 +279,65 @@ export class Recurly extends Emitter {
...
+   * @param {String} route
+   * @param {Object} data
+   * @param {Array} data.by must be an Array
+   * @param {String} by
+   * @param {Number} size=100
+   * @param {Function} done
+   * @private
+   */
+
+  pipedRequest ({ method, route, data, by, size }) {
```
snodgrass23: What about setting the default for `size` here like it appears in the doc block above. Since the only use of this function at this time using the default, we could remove the arg in that call. Is the `method` generally always going to be a `GET`? If so, we could set that default as well.

chrissrogers: I actually fixed the default assignment in #393 when I added specs for this method. I'll add it here too.
+1 to setting the method default as well

lib/recurly/apple-pay.js
chrissrogers: Yeah. Prior to writing the attachment, I'll adapt these modules to use a Pricing generic proxy

lib/recurly/pricing/subscription/index.js
```diff
@@ -334,6 +335,31 @@ export default class SubscriptionPricing extends Pricing {
...
+
+  /**
+   * Checks whether a given coupon may apply to this subscription
+   * @param  {Object} coupon
+   * @return {Boolean}
+   */
+  couponIsValidForSubscription (coupon) {
+    if (!coupon) return false;
+    if (!coupon.applies_to_plans) return false;
+    if (coupon.applies_to_all_plans) return true;
```
snodgrass23: another option would be to consolidate these last three lines into:

```
return coupon.applies_to_all_plans || ~coupon.plans.indexOf(this.items.plan.code)
```

chrissrogers: For business rules like this, I personally prefer the readability of the early returns

lib/recurly/apple-pay.js
chrissrogers: See #394

PR #386, opened 2017-11-02
Removes components

chrissrogers:
- Obsolete since #228


components/chrissrogers-noop@master/index.js
```diff
@@ -1 +0,0 @@
-module.exports = function () {};
```
chrissrogers: Best code I ever wrote

PR #392, opened 2017-11-06
add support for adyen hpp checkout

adeitrick:
Adds support for Adyen HPP Checkout


lib/recurly/adyen.js
```diff
@@ -0,0 +1,53 @@
+import Emitter from 'component-emitter';
+import errors from '../errors';
+import {Pricing} from './pricing';
+import {normalize} from '../util/normalize';
```
chrissrogers: Only Emitter is used amongst these imports

lib/recurly/adyen.js
```diff
@@ -0,0 +1,53 @@
...
+
+const debug = require('debug')('recurly:adyen');
```
chrissrogers: Perhaps add a few debug calls, on construct and start

lib/recurly/adyen.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+export function factory (options) {
+  return new Adyen(Object.assign({}, options, { recurly: this }));
+};
+
+
+/**
+ * Initializes an Adyen session.
+ *
+ * @param {Object} options
```
chrissrogers: Can you add the options? Looks like just `@param {Recurly} options.recurly`

lib/recurly/adyen.js
```diff
@@ -0,0 +1,53 @@
...
+  constructor (options) {
+    super();
+
+    this._ready = false;
+    this.config = {};
+    this.once('ready', () => this._ready = true);
+    this.recurly = options.recurly
+  }
+
+  start (opts) {
```
chrissrogers: Mind adding a jsdoc entry for the options?

lib/recurly/adyen.js
```diff
@@ -0,0 +1,53 @@
...
+      shopperLocale: opts.shopperLocale,
+      currencyCode: opts.currencyCode,
+      skinCode: opts.skinCode
+    };
+
+    const frame = this.recurly.Frame({ height: 600, path: '/adyen/start', payload });
+    frame.once('error', cause => this.error('paypal-tokenize-error', { cause }));
+
+    frame.once('done', function(token) {
+      this.emit('token', token)
```
chrissrogers: I suspect this needs to be in an arrow fn closure to emit properly

```
frame.once('done', token => this.emit('token', token));
```

lib/recurly/adyen.js
```diff
@@ -0,0 +1,53 @@
...
+ * @param {Object} options
+ * @constructor
+ * @public
+ */
+
+class Adyen extends Emitter {
+  constructor (options) {
+    super();
+
+    this._ready = false;
```
chrissrogers: `_ready` is never checked. Should it be on start?

lib/recurly/adyen.js
```diff
@@ -0,0 +1,53 @@
...
+    this.config = {};
```
chrissrogers: This isn't used

lib/recurly/adyen.js
```diff
@@ -0,0 +1,53 @@
...
+    const payload = {
+      invoiceUuid: opts.invoiceUuid,
+      countryCode: opts.countryCode,
+      shopperLocale: opts.shopperLocale,
+      currencyCode: opts.currencyCode,
+      skinCode: opts.skinCode
+    };
+
+    const frame = this.recurly.Frame({ height: 600, path: '/adyen/start', payload });
+    frame.once('error', cause => this.error('paypal-tokenize-error', { cause }));
```
chrissrogers: Do these errors cover instances of bad options, like an invalid countryCode, shopperLocal, etc. ? Wondering if we need some kind of validation on those values.

adeitrick: Adyen will do its own validations on their end when we try to invoke the modal, although there aren't great errors provided from them around validations. I suppose we could do validations around string length and whatnot here, but not sure how much that will buy us.

lib/recurly/adyen.js
adeitrick: not sure its necessary, going to remove

PR #396, opened 2017-11-10
Mobile Tabbing for Recurly.js

jpgnotgif:
#### Description
- These changes will allow customers to use the tab buttons in Safari on iOS to tab from form inputs to Recurly-js hosted field inputs

### DIFF
--- a/lib/recurly.js
+++ b/lib/recurly.js
@@ -134,7 +134,6 @@ export class Recurly extends Emitter {
 
   configure (options) {
     debug('configure');
-
     options = clone(options);
 
     if (typeof options === 'string') options = { publicKey: options };
@@ -174,8 +173,9 @@ export class Recurly extends Emitter {
     this.config.required = options.required || this.config.required || [];
 
     // Begin parent role configuration and setup
-    if (this.config.parent) this.parent();
-    else {
+    if (this.config.parent) {
+      this.parent();
+    } else {
       if (options.parentVersion) this.config.parentVersion = options.parentVersion;
     }
 
--- a/lib/recurly/hosted-field.js
+++ b/lib/recurly/hosted-field.js
@@ -3,6 +3,7 @@ import clone from 'component-clone';
 import dom from '../util/dom';
 import errors from '../errors';
 
+const bowser = require('bowser');
 const debug = require('debug')('recurly:hostedField');
 
 /**
@@ -83,7 +84,6 @@ export class HostedField extends Emitter {
   }
 
   // Private
-
   configure (options) {
     options = clone(options);
     this.target = dom.element(global.document.querySelector(options.selector));
@@ -95,25 +95,44 @@ export class HostedField extends Emitter {
   }
 
   inject () {
-    this.target.innerHTML = `
-      <div class="${this.classList}">
-        <iframe
-          src="${this.url}"
-          border="0"
-          frameborder="0"
-          allowtransparency="true"
-          scrolling="no">
-        </iframe>
-      </div>
-    `;
+    this.target.innerHTML = this.fieldHtml();
+    this.container = this.target.children[0]; // recurly-hosted-field
 
-    this.container = this.target.children[0];
-    this.iframe = this.container.children[0];
+    if (isMobile()) {
+      this.injectMobileElements();
+    } else {
+      this.iframe = this.container.children[0];
+    }
+
     this.window = this.iframe.contentWindow;
-
     this.iframe.style.height = '100%';
     this.iframe.style.width = '100%';
     this.iframe.style.background = 'transparent';
+  }
+
+  injectMobileElements() {
+    this.tabbingProxy = this.container.children[0];
+    this.iframe = this.container.children[1];
+    this.bindDeferredFocusOnTabbingProxy();
+  }
+
+  fieldHtml () {
+    return `<div class="${this.classList}">
+      ${isMobile() ? dom.createHiddenInput().outerHTML : ''}
+      <iframe
+        src="${this.url}"
+        border="0"
+        frameborder="0"
+        allowtransparency="true"
+        scrolling="no">
+      </iframe>
+      </div>
+    `;
+  }
+
+  bindDeferredFocusOnTabbingProxy () {
+    if (this.tabbingProxy === undefined) return;
+    this.tabbingProxy.addEventListener('focus', () => this.focus());
   }
 
   bindDeferredFocus () {
@@ -161,4 +180,13 @@ export class HostedField extends Emitter {
     if (!this.bus) return;
     this.bus.send(`hostedField:${this.type}:focus!`);
   }
+}
+
+/**
+ * Determine if user's browser is mobile or non-mobile
+ * @return {Boolean} value
+ */
+
+function isMobile() {
+  return bowser.mobile || bowser.tablet;
 }
--- a/lib/recurly/hosted-fields.js
+++ b/lib/recurly/hosted-fields.js
@@ -1,5 +1,6 @@
 import clone from 'component-clone';
 import Emitter from 'component-emitter';
+import tabbable from 'tabbable';
 import {HostedField} from './hosted-field';
 import deepAssign from '../util/deep-assign';
 import errors from '../errors';
@@ -29,6 +30,8 @@ export class HostedFields extends Emitter {
     this.readyState = 0;
     this.configure(options);
     this.inject();
+    this.on('hostedField:tab:previous', this.onTab.bind(this, 'previous'));
+    this.on('hostedField:tab:next', this.onTab.bind(this, 'next'));
     this.on('hostedField:state:change', this.update.bind(this));
     this.on('bus:added', bus => {
       this.bus = bus;
@@ -63,7 +66,7 @@ export class HostedFields extends Emitter {
   }
 
   inject () {
-    this.on('hostedField:ready', this.readyHandler.bind(this));
+    this.on('hostedField:ready', this.onReady.bind(this));
     FIELD_TYPES.forEach(type => {
       try {
         this.fields.push(new HostedField(this.fieldConfig(type)));
@@ -109,15 +112,36 @@ export class HostedFields extends Emitter {
     this.initQueue = [];
   }
 
-  readyHandler (body) {
+  onReady (body) {
     const pos = this.initQueue.indexOf(body.type);
     if (~pos) this.initQueue.splice(pos, 1);
     if (this.initQueue.length === 0) {
-      this.off('hostedField:ready', this.readyHandler);
+      this.off('hostedField:ready', this.onReady);
       this.bus.send('hostedFields:ready');
       this.ready = true;
     }
     this.update(body);
+  }
+
+  onTab (direction, message) {
+    const origin = this.getFieldByType(message.type);
+    let tabbableItems = this.tabbableItems();
+    let destination;
+
+    // Find the origin within the tabbable list, and focus in our intended direction
+    tabbableItems.forEach((tabbableItem, i) => {
+      if (origin instanceof HostedField) {
+        if (origin.tabbingProxy == tabbableItem) {
+          destination = direction == 'previous' ? tabbableItems[i - 1] : tabbableItems[i + 1];
+        }
+      }
+    });
+
+    if (destination) destination.focus();
+  }
+
+  tabbableItems () {
+    return tabbable(global.document.body);
   }
 
   update (body) {
@@ -139,5 +163,9 @@ export class HostedFields extends Emitter {
       style: deepAssign({}, fields.all.style, fields[type].style),
       recurly: this.config.recurly
     };
+  }
+
+  getFieldByType (type) {
+    return this.fields.filter(f => f.config.type === type)[0];
   }
 }
--- a/lib/util/dom.js
+++ b/lib/util/dom.js
@@ -10,10 +10,11 @@ var each = require('component-each');
  */
 
 module.exports = {
-  element: element,
-  value: value,
+  createHiddenInput: createHiddenInput,
   data: data,
-  findNodeInParents: findNodeInParents
+  element: element,
+  findNodeInParents: findNodeInParents,
+  value: value
 };
 
 /**
@@ -185,4 +186,18 @@ function findNodeInParents(node, targetTagName) {
   if (!node) return null;
   if (node.tagName.match(RegExp(targetTagName, 'i'))) return node;
   return findNodeInParents(node.parentNode, targetTagName);
+}
+
+/**
+ * Utility function to create a hidden input element
+ * @return {InputElement}
+ */
+function createHiddenInput (attributes = {}) {
+  let hidden = global.document.createElement(attributes.type === 'button' ? 'button' : 'input');
+  if (!('type' in attributes)) attributes.type = 'text';
+  if (!('style' in attributes)) attributes.style = 'position: absolute; top: 0px; left: -1000px; opacity: 0;';
+
+  Object.keys(attributes).forEach(attr => hidden.setAttribute(attr, attributes[attr]));
+
+  return hidden;
 }
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "babel-loader": "^7.1.2",
     "babel-plugin-transform-object-assign": "^6.22.0",
     "babel-preset-es2015": "^6.24.0",
+    "bowser": "^1.9.0",
     "combinations": "^0.1.1",
     "css-loader": "^0.28.7",
     "extract-text-webpack-plugin": "^3.0.2",
@@ -77,6 +78,7 @@
     "phantomjs-prebuilt": "^2.1.14",
     "sinon": "^2.1.0",
     "style-loader": "^0.19.0",
+    "tabbable": "^1.1.2",
     "webpack": "^3.9.1",
     "webpack-dev-server": "^2.9.5"
   }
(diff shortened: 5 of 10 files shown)


lib/recurly.js
```diff
@@ -119,8 +119,6 @@ export class Recurly extends Emitter {
    */
 
   configure (options) {
-    debug('configure');
```
chrissrogers: Let's keep this call in

lib/recurly/hosted-field.js
```diff
@@ -85,20 +87,33 @@ export class HostedField extends Emitter {
...
+
+    this.hiddenInputProxy = dom.createHiddenInput();
+    this.hiddenInputProxy.addEventListener('focus', () => {
+      this.focus();
+    });
+
+    this.target.innerHTML = hostedDiv;
+
+    this.target.children[0].insertAdjacentElement('afterbegin', this.hiddenInputProxy);
+    this.target.children[0].children[0].insertAdjacentHTML('afterend', hostedIframe);
```
chrissrogers: Instead of building and inserting this separately from the html text above, let's just write it into the `this.target.innerHTML` assignment, then find it in the injected DOM structure as we do for `this.container`, `this.iframe`, etc. As this is written, it's difficult to tell where the input proxy is meant to reside in the tree.

lib/recurly/hosted-field.js
```diff
@@ -148,4 +163,9 @@ export class HostedField extends Emitter {
     if (!this.bus) return;
     this.bus.send(`hostedField:${this.type}:focus!`);
   }
+
+  // TODO: Move this out into a generic message factory
+  message (name, body = {}) {
+    global.parent.postMessage(new Message(`hostedField:${name}`, merge(true, body, { type: this.config.type })), '*');
+  }
```
chrissrogers: This isn't used and doesn't function in the parent.

lib/recurly/hosted-fields.js
```diff
@@ -96,17 +101,94 @@ export class HostedFields extends Emitter {
...
+          let nextEl = nextItem instanceof HostedField ? nextItem.container : nextItem;
+          if (!dom.element(el)) return;
+          if (!dom.element(nextEl)) return;
+          if (!(field.container.compareDocumentPosition(el) & Node.DOCUMENT_POSITION_PRECEDING)) return;
+          if (!(field.container.compareDocumentPosition(nextEl) & Node.DOCUMENT_POSITION_FOLLOWING)) return;
+          tabbableItems.splice(i + 1, 0, field);
+          return true;
+        });
+      }
+    });
```
chrissrogers: I don't think this is necessary since there are input proxies being injected before the hosted field iframes.

lib/recurly/hosted-fields.js
```diff
@@ -96,17 +101,94 @@ export class HostedFields extends Emitter {
...
+          }
+          if (destination === undefined) {
+            fieldsSlice.find((field, fieldIndex) => {
+              if (!(dom.isHiddenInput(field))) {
+                destination = field;
+                return destination;
+              }
+            });
+          }
+      }
```
chrissrogers: There's a bit of repetition here, and the intent of the fieldSlice isn't clear. I have to believe the nested looping shouldn't be needed either, since we know our origin and direction within the tabbable set when onTab is called.

lib/recurly/hosted-fields.js
```diff
@@ -96,17 +101,94 @@ export class HostedFields extends Emitter {
...
+
+      destination.focus();
+      return true;
```
chrissrogers: Shouldn't need to return here

lib/recurly/hosted-fields.js
jpgnotgif: @chrissrogers we talked about this previously but I'd like to re-emphasize the issue again.

I attempted to undo the above changes and rely strictly on the hidden inputs to determine the adjacent iframe input for focus. However, in trying to reference the input within the iframe, I was receiving the infamous `DOM Exception 18` which does not permit access to elements within an iframe if the iframe src and its parent src differ.

chrissrogers: Per discussion, the focus intent should be implemented by message bus and not through direct DOM access. 👍

lib/recurly/hosted-field.js
```diff
@@ -85,21 +87,21 @@ export class HostedField extends Emitter {
...
+    this.target.innerHTML = dom.hostedFieldHtml(this.classList, this.url);
+
+    if (dom.isMobile()) {
+      this.container = this.target.children[0]; // recurly-hosted-field
+      this.tabbingProxy = this.container.children[0];
+      this.iframe = this.container.children[1]; // iframe
+      this.window = this.iframe.contentWindow;
+      this.tabbingProxy.addEventListener('focus', () => {
+        this.focus();
+      });
```
chrissrogers: Since this method is more concerned with injecting into the DOM, this should move out to its own method akin to `bindDeferredFocus`

lib/recurly/hosted-fields.js
```diff
@@ -1,5 +1,7 @@
 import clone from 'component-clone';
 import Emitter from 'component-emitter';
+import dom from '../util/dom';
```
chrissrogers: This isn't used

lib/recurly/hosted-fields.js
```diff
@@ -96,17 +100,37 @@ export class HostedFields extends Emitter {
...
+        if (origin.tabbingProxy === el) {
+          destination = direction === 'previous' ? tabbableItems[i - 1] : tabbableItems[i + 1];
+        }
+      } else {
+          destination = direction === 'previous' ? (tabbableItems[i - 1] || last(tabbableItems)) : (tabbableItems[i + 1] || tabbableItems[0]);
+      }
+    });
+
+    if (destination !== undefined)
+      destination.focus();
```
chrissrogers: nit:
```js
if (destination) destination.focus();
```

lib/recurly/hosted-fields.js
```diff
@@ -96,17 +100,37 @@ export class HostedFields extends Emitter {
...
+    let destination;
+
+    // Find the origin within the tabbable list, and focus in our intended direction
+    tabbableItems.find((el, i) => {
+      if (origin instanceof HostedField) {
+        if (origin.tabbingProxy === el) {
+          destination = direction === 'previous' ? tabbableItems[i - 1] : tabbableItems[i + 1];
+        }
+      } else {
+          destination = direction === 'previous' ? (tabbableItems[i - 1] || last(tabbableItems)) : (tabbableItems[i + 1] || tabbableItems[0]);
```
chrissrogers: is this logic path ever hit? `getFieldByType` will always return a `HostedField`, and this line references a function `last` that is never defined.

lib/recurly/hosted-fields.js
```diff
@@ -126,4 +150,8 @@ export class HostedFields extends Emitter {
       recurly: this.config.recurly
     };
   }
+
+  getFieldByType (type) {
+    return this.fields.find(field => field.config.type === type);
```
chrissrogers: `Array.prototype.find` isn't available in IE

lib/util/dom.js
```diff
@@ -9,11 +9,17 @@ var each = require('component-each');
...
 module.exports = {
-  element: element,
-  value: value,
+  createHiddenInput: createHiddenInput,
   data: data,
-  findNodeInParents: findNodeInParents
+  element: element,
+  findNodeInParents: findNodeInParents,
+  hostedFieldHtml: hostedFieldHtml,
+  isHiddenInput: isHiddenInput,
```
chrissrogers: This is never used

lib/util/dom.js
```diff
@@ -186,3 +192,74 @@ function findNodeInParents(node, targetTagName) {
...
+ * Utility function to generate injectable html for a hosted field
+ * @param  {String} classList
+ * @params {String} url
+ * @return {String} html
+ */
+
+function hostedFieldHtml (classList = [], url = '') {
+  let html;
+
+  if (this.isMobile()) {
```
chrissrogers: This reference isn't stable. It only works because it's on the module export

lib/util/dom.js
```diff
@@ -186,3 +192,74 @@ function findNodeInParents(node, targetTagName) {
...
+}
+
+/**
+ * Determine if user's browser is mobile or non-mobile
+ * @return {Boolean} value
+ */
+
+function isMobile() {
+  return bowser.mobile || bowser.tablet;
+}
```
chrissrogers: This doesn't really fit the scope of the dom util since it doesn't interact with the dom.

lib/util/dom.js
```diff
@@ -186,3 +192,74 @@ function findNodeInParents(node, targetTagName) {
...
+        <div class="${classList}">
+          <iframe
+            src="${url}"
+            border="0"
+            frameborder="0"
+            allowtransparency="true"
+            scrolling="no">
+          </iframe>
+        </div>
+      `;
```
chrissrogers: It would be better to not repeat the iframe string in the two condition branches.

```js
return `<div class="${classList}">
  {isMobile ? this.createHiddenInput().outerHTML : ''}
  <iframe
    src="${url}"
    border="0"
    frameborder="0"
    allowtransparency="true"
    scrolling="no">
  </iframe>
</div>
`
```

package.json
```diff
@@ -24,9 +25,11 @@
     "lodash.isempty": "^4.4.0",
     "lodash.isfinite": "^3.3.2",
     "lodash.pick": "^4.4.0",
+    "merge": "^1.2.0",
```
chrissrogers: Let's keep this out

test/hosted-fields.test.js
```diff
@@ -0,0 +1,68 @@
...
+  const hostedField = hostedFields.fields[0];
+
+  describe('initialization', () => {
+    it('builds fields collection', () => {
+      for (let i = 0; i < hostedFields.fields.length; i++) {
+        assert(hostedFields.fields[i]);
+        FIELD_TYPES.forEach(fieldType => {
+          if (fieldType == hostedFields.fields[i].type)
+            assert.equal(fieldType, hostedFields.fields[i].type);
+        });
```
chrissrogers: I think this can be written more simply as

```js
assert(~FIELD_TYPES.indexOf(hostedFields.fields[i].type));
```

test/hosted-fields.test.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  fixture('all', data);
+  const recurly = initRecurly();
+  const hostedFields = new HostedFields({ recurly: recurly.config });
+  const hostedField = hostedFields.fields[0];
+
+  describe('initialization', () => {
+    it('builds fields collection', () => {
+      for (let i = 0; i < hostedFields.fields.length; i++) {
```
chrissrogers: `hostedFields.fields` is an Array, so you can use forEach here

test/hosted-fields.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+  describe('tabbing', () => {
+    it('binds to specific events', () => {
+      const events = [
+        'bus:added',
+        'hostedField:state:change',
+        'hostedField:tab:next',
+        'hostedField:tab:previous'
+      ]
+      for (let i = 0; i < events.length; i++) {
```
chrissrogers: `forEach` is nicer here

lib/recurly/hosted-fields.js
```diff
@@ -109,17 +112,38 @@ export class HostedFields extends Emitter {
...
+        if (origin.tabbingProxy == tabbableItems[i]) {
+          destination = direction == 'previous' ? tabbableItems[i - 1] : tabbableItems[i + 1];
+        }
+      }
+    }
+
+    if (destination) destination.focus();
+  }
+
+  tabbableItems () {
```
jpgnotgif: @chrissrogers : I had to wrap the tabbable call in order to properly spy on it. See below:

https://github.com/sinonjs/sinon/issues/664

test/server/fixtures/field.html.ejs
```diff
@@ -26,7 +25,7 @@
 
       for (var i = 0; i < window.parent.frames.length; i++) {
         var frame = window.parent.frames[i].contentWindow || window.parent.frames[i];
-        inputs[frame.config().type] = frame.value();
+        if (frame !== undefined) inputs[frame.config().type] = frame.value();
```
jpgnotgif: @chrissrogers : this is how I deal with the stale relays

chrissrogers: I think these frames shouldn't be ignored here, and the fact that they're causing issues indicates that they should be present.

lib/recurly/hosted-fields.js
```diff
@@ -140,4 +164,12 @@ export class HostedFields extends Emitter {
       recurly: this.config.recurly
     };
   }
+
+  getFieldByType (type) {
+    let field;
+    for (let i = 0; i < this.fields.length; i++) {
+      field = this.fields[i];
+      if (field.config.type == type) return field;
+    }
```
chrissrogers: This would be more clean as

```js
this.fields.filter(f => f.config.type === type)[0];
```

test/configure.test.js
```diff
@@ -240,6 +240,7 @@ describe('Recurly.configure', function () {
       configureRecurly(this.recurly, 1, () => {
         assert(document.querySelector('#number-1 iframe') instanceof HTMLIFrameElement);
         assert.equal(document.querySelector('#number-2 iframe'), null);
+        done();
```
chrissrogers: is this errant?

test/hosted-field.test.js
```diff
@@ -0,0 +1,42 @@
+import assert from 'assert';
+import bowser from 'bowser';
+
+import { fixture } from './support/fixtures';
+import { initRecurly } from './support/helpers';
+import { HostedField } from '../lib/recurly/hosted-field';
+import { Recurly } from '../lib/recurly';
```
chrissrogers: nit: spacing

test/hosted-field.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+describe('HostedField', () => {
+  fixture('minimal');
```
chrissrogers: Instead of calling `fixture` here, let's use `applyFixtures` and set `this.ctx.fixture` in order to have it built and torn down per test.

test/hosted-field.test.js
```diff
@@ -0,0 +1,42 @@
...
+  const recurly = initRecurly();
+
+  const fieldConfig = {
+    type: fieldType,
+    selector: `[data-recurly=${fieldType}]`,
+    recurly: recurly.config
+  };
+
+  describe('on mobile', () => {
+    sinon.stub(bowser, 'mobile').value(true);
```
chrissrogers: 👍

test/hosted-fields.test.js
```diff
@@ -0,0 +1,83 @@
+import assert from 'assert';
+import bowser from 'bowser';
+import tabbable from 'tabbable';
+
+import { applyFixtures } from './support/fixtures';
+import { initRecurly, nextTick } from './support/helpers';
+import HostedField from '../lib/recurly/hosted-field';
+import { FIELD_TYPES, HostedFields } from '../lib/recurly/hosted-fields';
+import { Recurly } from '../lib/recurly';
```
chrissrogers: nit: spacing

test/server/fixtures/field.html.ejs
```diff
@@ -23,9 +22,15 @@
       var recurly = new parent.recurly.Recurly(recurlyConfig());
       var inputs = body.inputs;
       var id = body.id;
+      var frame;
+
+      for (var i = 0; i < window.parent.frames.length; i++) {
+        frame = window.parent.frames[i].contentWindow || window.parent.frames[i];
+        if (frame === undefined) window.parent.frames.splice(i, 1);
+      }
```
chrissrogers: Is the intent of this to clear a relay frame that's been abandoned? If so, the spec causing it needs to be fixed so that this isn't necessary

lib/recurly/hosted-fields.js
```diff
@@ -109,17 +112,38 @@ export class HostedFields extends Emitter {
...
     this.update(body);
   }
 
+  onTab (direction, message) {
+    const origin = this.getFieldByType(message.type);
+    let tabbableItems = this.tabbableItems();
+    let destination;
+
+    // Find the origin within the tabbable list, and focus in our intended direction
+    for (let i = 0; i < tabbableItems.length; i++) {
```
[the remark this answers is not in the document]
jpgnotgif: `tabbableItems` is an `Array`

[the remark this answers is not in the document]
chrissrogers: Ah okay, then we can use forEach here to clean it up just a bit

test/server/fixtures/field.html.ejs
jpgnotgif: I'll narrow down the spec that's causing this to break, update the spec and update this .ejs file

lib/recurly/hosted-field.js
```diff
@@ -95,27 +95,46 @@ export class HostedField extends Emitter {
...
     this.iframe.style.background = 'transparent';
   }
 
+  injectMobileElements() {
+    this.tabbingProxy = this.container.children[0];
+    this.iframe = this.container.children[1];
+    this.bindDeferredFocusOnTabbingProxy();
+  }
+
+  fieldHtml (classList = [], url = '') {
```
chrissrogers: Don't need to pass these in since they're on `this`

PR #404, opened 2017-11-30
Adds Card Hosted Field support

chrissrogers:
- Adds `recurly-hosted-field-valid` css class
- Adds `recurly-hosted-field-invalid` css class
- Adds `recurly-hosted-field-[type]-valid` css class
- Adds `recurly-hosted-field-[type]-invalid` css class
- Adapts hosted field config validation to support a single card hosted field

### DIFF
--- a/lib/recurly.js
+++ b/lib/recurly.js
@@ -74,6 +74,10 @@ const defaults = {
     cvv: {
       selector: '[data-recurly=cvv]',
       style: {}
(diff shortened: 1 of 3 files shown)


chrissrogers:
@snodgrass23 the push CI run fails because the tag has since been deleted -- CI run acceptance should be based on the PR run, which has succeeded

PR #406, opened 2017-12-02
Removes lodash.after dependency from BraintreeStrategy

chrissrogers:
- This is targeting `v4.8.0`
- Further reduces dist size


snodgrass23:
Do we want to remove `lodash.after` from the `package.json` file too?

chrissrogers:
It's actually already been removed, but webpack was able to resolve the dependency anyway (I think because it's a devDependency)

PR #412, opened 2017-12-05
fix bug with listener not registered for cancel event on paypal

keemy:
Currently listeners registered to the cancel event don't get called for paypal/braintree. This fix makes them get called.

### DIFF
--- a/lib/recurly/paypal/index.js
+++ b/lib/recurly/paypal/index.js
@@ -18,7 +18,8 @@ export function factory (options) {
 const DEFERRED_EVENTS = [
   'ready',
   'token',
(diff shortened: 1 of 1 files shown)


chrissrogers:
Good catch!

chrissrogers:
This will be pulled into v4.8.0

keemy:
@chrissrogers cool thanks. Is there an eta on when v4.8.0 will be out? Just want to see if I should run the fork or if it's soon I'll just wait.

chrissrogers:
We are targeting next week, but that could be held up. I'll let you know here if that's the case.

PR #420, opened 2018-01-16
Revert "Mobile Tabbing for Recurly.js"

chrissrogers:
- Reverts recurly/recurly-js#396
- Support for the combined card field is needed, and an interim release to include other features is pending


jpgnotgif:
👍

PR #421, opened 2018-01-19
Mobile Tabbing for multiple/combined card fields

jpgnotgif:
#### Description
- These changes will allow customers to use the tab buttons in Safari on iOS to tab from form inputs to Recurly-js hosted field inputs. Note that the hosted field inputs may be multiple inputs per field or a single combined field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Recurly.js CHANGELOG
 
-### Version 4.8.2 (Feb 15,2018)
+### Version 4.8.2 (Feb 15, 2018)
 
 * Adds support for Apple pay zero amount authorizations [#426][426]
 
--- a/lib/recurly/hosted-field.js
+++ b/lib/recurly/hosted-field.js
@@ -3,6 +3,7 @@ import clone from 'component-clone';
 import dom from '../util/dom';
 import errors from '../errors';
 
+const bowser = require('bowser');
 const debug = require('debug')('recurly:hostedField');
 
 /**
@@ -83,7 +84,6 @@ export class HostedField extends Emitter {
   }
 
   // Private
-
   configure (options) {
     options = clone(options);
     this.target = dom.element(global.document.querySelector(options.selector));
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 ## Recurly.js CHANGELOG
 
-### Version 4.8.1 (Jan 16,2018)
+### Version 4.8.1 (Jan 5,2018)
 
-* Adds support for setting PayPal logo in Express Checkout flow [#418][418]
+* Adds support for tabbing between hosted fields on mobile devices [#396][396]
```
chrissrogers: This diff is out of date

lib/recurly/hosted-fields.js
```diff
@@ -63,7 +67,17 @@ export class HostedFields extends Emitter {
...
+    this.on('hostedField:ready', this.onReady.bind(this));
+
+    if (this.serveAsSingleField()) {
+      const fieldType = 'card';
+      const fieldConfig = this.fieldConfig(fieldType);
+      Object.assign(fieldConfig, { singleField: true });
+      this.fields.push(new HostedField(fieldConfig));
+      this.initQueue.push(fieldType);
+      return;
+    }
```
chrissrogers: There are far too many assumptions here about field type and I don't like that the fieldConfig is being mutated here. It should really be static from the time it is read from `this.fieldConfig`, else that method is incomplete. This routine is also somewhat redundant given the iterator below.

lib/recurly/hosted-fields.js
```diff
@@ -140,4 +175,24 @@ export class HostedFields extends Emitter {
...
+    FIELD_TYPES.forEach((type, _) => {
+      const selector = fields[type].selector;
+      const element = dom.element(global.document.querySelector(selector));
+      element ? elementCounts['found']++ : elementCounts['notFound']++;
+    });
+    /* Verify that at least one element is NOT in the dom. If at least one element
+     * is not in the dom, then we will configure a single hosted field. */
+    return elementCounts['notFound'] > 1;
+  }
```
chrissrogers: I don't think this method is necessary, and I don't think single fields need to be treated differently than others. No matter if there are one or many hosted fields being configured, each should get a hidden tabbing sink, no?

lib/recurly/hosted-field.js
```diff
@@ -95,27 +95,48 @@ export class HostedField extends Emitter {
...
+      ${isMobile ? dom.createHiddenInput().outerHTML : ''}
+      <iframe
+        src="${this.url}"
+        border="0"
+        frameborder="0"
+        allowtransparency="true"
+        scrolling="no">
+      </iframe>
+      </div>
+      ${isMobile && this.config.singleField ? dom.createHiddenInput().outerHTML : ''}
```
chrissrogers: I don't think it's necessary for a singular hosted field needs any special treatment here. It should be treated the same as if there were multiple, no?

chrissrogers: A bit of a nitpick, but would you mind going back to the former nesting pattern in this html template? It's a bit easier to read.

lib/recurly/hosted-fields.js
chrissrogers: I think it's ultimately just not necessary to detect whether there are one or many hosted fields, since they should all get just one tabbing sink :)

PR #426, opened 2018-02-05
Apple pay zero amount authorizations

chrissrogers:
- Apple Pay cannot process zero amount transactions
- When a transaction is discounted to zero, we will authorize the card with a temporary one-unit-amount transaction
- This updates `recurly.ApplePay` to reflect this temporary authorization when the total has been reduced to zero via discount, free trial, or simply base price

##### Customizing the authorization message
```js
var applePay = recurly.ApplePay({
  // ...
  i18n: {
    authorizationLineItemLabel: 'Temporary Card Authorization',
    subtotalLineItemLabel: 'Subtotal',
    discountLineItemLabel: 'Discount',
    taxLineItemLabel: 'Tax',
    giftCardLineItemLabel: 'Gift card'
  }
});
```

### DIFF
--- a/lib/recurly/apple-pay.js
+++ b/lib/recurly/apple-pay.js
@@ -16,7 +16,11 @@ const APPLE_PAY_ADDRESS_MAP = {
   state: 'administrativeArea',
   postal_code: 'postalCode',
   country: 'countryCode'
-}
+};
+
+const I18N = {
+  authorizationLineItemLabel: 'Card Authorization (Temporary)'
+};
 
 /**
  * Instantiation factory
@@ -48,7 +52,9 @@ class ApplePay extends Emitter {
     super();
 
     this._ready = false;
-    this.config = {};
+    this.config = {
+      i18n: I18N
+    };
     this.once('ready', () => this._ready = true);
 
     // Detect whether Apple Pay is available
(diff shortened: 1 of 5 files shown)


lib/recurly/apple-pay.js
```diff
@@ -96,15 +109,33 @@ class ApplePay extends Emitter {
    */
   get lineItems () {
     // Clone configured line items
-    return [].concat(this.config.lineItems);
+    const lineItems = [].concat(this.config.lineItems);
+    return lineItems;
```
snodgrass23: what's the benefit of assigning the value instead of directly returning it as before?

lib/recurly/apple-pay.js
```diff
@@ -240,7 +280,8 @@ class ApplePay extends Emitter {
    */
   onPaymentMethodSelected (event) {
     debug('Payment method selected', event);
-    this.session.completePaymentMethodSelection(this.total, this.lineItems);
+    console.log(this.lineItems);
```
snodgrass23: want to keep this log in here?

lib/recurly/apple-pay.js
chrissrogers: No benefit. This was probably an artifact of some routine I'd added and then subsequently removed here.

lib/recurly/apple-pay.js
chrissrogers: Nope! good catch

PR #433, opened 2018-03-13
Adds displayIcon field configuration option pass-through

chrissrogers:
- Adds a new `displayIcon` configuration value, which will inform a card field to not display its integrated card brand icon

#### Example

```js
recurly.configure({
  // ...
  fields: {
    card: {
      displayIcon: false
    }
  }
  // ...
});
```


snodgrass23:
Is this going to pass through to hosted fields for the actual logic to not display the icon?

PR #455, opened 2018-04-24
chore: remove eval from minified file

thisislawatts:
We wish to use `recurly.js` on a site which has a [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) in place that does not permit `unsafe-eval` for `script-src` resources.

> 'unsafe-eval'
    Allows the use of eval() and similar methods for creating code from strings. You must include the single quotes.

I believe that this is being introduced to your compiled codebase by Webpack. This thread has some further insight into people's experience here [#5627 Calling Function or eval should be eliminated as a security measure and would cause the application execution to be halted if the document CSP header has script-src 'self'; ](https://github.com/webpack/webpack/issues/5627)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,6 +3,9 @@ var path = require('path');
 var minify = ~process.argv.indexOf('-p');
 
 module.exports = {
+  node: {
+    global: false,
+  },
   entry: './index',
   output: {
     path: path.join(__dirname, 'build'),


chrissrogers:
I think it's a good idea to remove this; however, the build is currently broken by this change, since there are many references to `global`. Those will all need to be changed to reference `window`.

Linking to the relevant buildin: https://github.com/webpack/webpack/blob/master/buildin/global.js

thisislawatts:
Thanks for the quick response @chrissrogers. That certainly makes a lot of sense, I hadn't meant to leave a broken build for you to fix. Let me take a look into this.

chrissrogers:
👍

chrissrogers:
It appears on of our dependencies needs to be changed in order to comply with the `global` availability. I can handle that in short order and we should have this implemented by the next version.

chrissrogers:
#458 will fix this

PR #456, opened 2018-04-25
fix: switch to using window over global

thisislawatts:
As mentioned in #455 it would be good to remove use of eval. To facilitate that change we need to remove use of `global` within the library and instead use `window`

### DIFF
--- a/lib/recurly/apple-pay.js
+++ b/lib/recurly/apple-pay.js
@@ -63,9 +63,9 @@ class ApplePay extends Emitter {
     this.once('ready', () => this._ready = true);
 
     // Detect whether Apple Pay is available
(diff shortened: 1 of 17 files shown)


thisislawatts:
🕵️‍♀️

chrissrogers:
Thanks for this! Yeah that's exactly right on the sauce front. I'll just run the tests locally to confirm.

thisislawatts:
Thanks @chrissrogers let me know if there's anything else I can look into on this.

chrissrogers:
This is looking good. Thanks for the contrib!

thisislawatts:
@chrissrogers No problem, thanks for helping get this turned around so quickly. 🙇

PR #459, opened 2018-04-27
Recurly.Event

chrissrogers:
- Base branch for `Recurly.event` feature


lib/recurly/event.js
```diff
@@ -0,0 +1,35 @@
...
+    this.recurly.request.queued({ method: 'post', route: '/events', data });
+  }
+
+  /**
+   * Formats an event payload
+   *
+   * @param {Object} options
+   * @param {String} options.name event name
+   */
+  payload ({ name }) {
```
snodgrass23: Are you planning on using other properties in the future? If not, probably easier to just accept a single string argument instead of passing in the whole object and pulling a single prop.

chrissrogers: Yes exactly, the payload here will accept more properties as they are introduced. Those will come in an upcoming PR to this branch.

PR #463, opened 2018-06-04
Error reporting

chrissrogers:
- Adds event reporting for certain errors, to help us monitor occurrences


lib/recurly/errors.js
```diff
@@ -0,0 +1,298 @@
...
+  },
+  {
+    code: 'missing-hosted-field-target',
+    message: c => `Target element not found for ${c.type} field using selector '${c.selector}'`,
+    classification: 'merchant'
+  },
+  {
+    code: 'api-error',
+    message: 'There was an error with your request.',
+    classification: 'metal'
```
jpgnotgif: This is specific to Rails. Would something along the lines of `api`, `middleware`, or perhaps `service-layer` work?

chrissrogers: Sure, I'd agree that it's a bit overloaded. I think `api` is appropriate.

PR #466, opened 2018-06-09
Pricing reporting

chrissrogers:
- Built atop #465 
- Dispatches events when pricing changes occur


lib/recurly/pricing/index.js
```diff
@@ -124,6 +126,27 @@ export class Pricing extends Emitter {
     }, this).nodeify(done);
   }
 
+  /**
+   * Binds important events to the EventDispatcher
+   *
+   * @protected
+   */
+  bindDispatchedEvents () {
```
jpgnotgif: Would it make sense to constantize the event names?

PR #467, opened 2018-06-19
Report metadata

chrissrogers:
- Built atop #466 
- Submits event metadata

#### Bonus
- Quiets coupon unset events by fixing their emission case
- Ensures objects returned and emitted by pricing instances cannot mutate the pricing instances themselves


lib/recurly/pricing/subscription/index.js
```diff
@@ -19,6 +20,7 @@ export default class SubscriptionPricing extends Pricing {
     super(recurly);
     this.id = id;
     this.debug = debug;
+    this.recurly.report('pricing:subscription:create');
```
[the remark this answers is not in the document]
chrissrogers: It's okay here since the domain will always be 'subscription' given the class

lib/recurly/pricing/index.js
```diff
@@ -116,14 +123,54 @@ export class Pricing extends Emitter {
...
     }, this).nodeify(done);
   }
 
+  /**
+   * Utility to emit an event and call a PricingPromise resolver
+   * with a mutation-safe copy of the item object
+   *
+   * @param {string} event event name
+   * @param {Object} object pricing item
+   * @param {Functioon} resolve PricingPromise resolver
```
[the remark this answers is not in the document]
chrissrogers: 👍

lib/recurly/pricing/checkout/index.js
```diff
@@ -23,6 +24,7 @@ export default class CheckoutPricing extends Pricing {
   constructor (...args) {
     super(...args);
     this.debug = debug;
+    this.recurly.report('pricing:checkout:create');
```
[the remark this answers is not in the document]
chrissrogers: I do not see the comment referenced -- was it deleted? I think since this string is used only once it's fine for it to remain inline here.

lib/recurly.js
```diff
@@ -18,7 +19,7 @@ import {factory as Frame} from './recurly/frame';
 import {factory as PayPal} from './recurly/paypal';
 import {deprecated as deprecatedPaypal} from './recurly/paypal/strategy/direct';
 import {Bus} from './recurly/bus';
-import {EventDispatcher} from './recurly/event';
+import {Reporter} from './recurly/reporter';
 import {Fraud} from './recurly/fraud';
 import {HostedFields, FIELD_TYPES} from './recurly/hosted-fields';
 import {Request} from './recurly/request';
```
gla7: Do we want to order imports alphabetically or otherwise?

lib/recurly.js
```diff
@@ -289,6 +291,33 @@ export class Recurly extends Emitter {
...
+  }
+
+  /**
+   * Binds important events to the EventDispatcher
+   *
+   * @private
+   */
+  bindReporting () {
+    if (!this.isParent) return;
+    ['focus', 'blur'].forEach(eventName => {
```
gla7: Perhaps we can include these as a constant e.g. `EVENTS = ['focus', 'blur']`?

lib/recurly.js
```diff
@@ -289,6 +291,33 @@ export class Recurly extends Emitter {
...
+      this.on(`hostedField:${eventName}`, ({ type }) => {
+        const state = this.hostedFields.state[type];
+        let meta = pick(state, ['valid', 'empty']);
```
gla7: I think `meta` should be a `const`, as you are later adding an attribute to the object, but the reference object is the same.

lib/recurly/errors.js
```diff
@@ -272,13 +272,13 @@ function recurlyErrorFactory (definition) {
       this.message += ` (need help? ${this.help})`;
     }
 
-    if (this.reporter instanceof EventDispatcher) {
+    if (this.reporter instanceof Reporter) {
       let type = 'client';
       if (this.classification) type += `:${this.classification}`;
 
       // Warning: any errors that occur in this code path risk
       // a stack overflow if they include a reporter
```
gla7: What is done to prevent said stack overflow?

lib/recurly/pricing/checkout/index.js
```diff
@@ -238,47 +234,47 @@ export default class CheckoutPricing extends Pricing {
...
    * @return {PricingPromise}
    * @public
    */
   coupon (couponCode) {
-    const unset = () => {
-      debug('unset.coupon');
-      delete this.items.coupon;
-      this.emit('unset.coupon');
-    };
+    if (~this.couponCodes.indexOf(couponCode)) {
```
gla7: Curious: what is this doing? Could we assign this to a new variable that describes what this does?

lib/recurly/pricing/subscription/index.js
```diff
@@ -84,25 +86,21 @@ export default class SubscriptionPricing extends Pricing {
     return new PricingPromise((resolve, reject) => {
       if (currentPlan && currentPlan.code === planCode) {
         currentPlan.quantity = quantity;
-        return resolve(currentPlan);
+        return resolve(clone(currentPlan));
```
gla7: Why is it necessary to make a deep copy here?

lib/recurly/pricing/subscription/index.js
```diff
@@ -213,40 +201,42 @@ export default class SubscriptionPricing extends Pricing {
...
    * @param {Function} [done] callback
    * @public
    */
   coupon (newCoupon, done) {
-    const unsetCoupon = () => {
-      debug('unset.coupon');
-      delete this.items.coupon;
-      this.emit('unset.coupon');
-    };
+    if (~this.couponCodes.indexOf(newCoupon)) {
```
gla7: Same question as above ^

lib/recurly/pricing/checkout/index.js
chrissrogers: It's a common way of detecting if an element is in an Array

lib/recurly/errors.js
chrissrogers: Not including a reporter call.

lib/recurly.js
```diff
@@ -230,7 +232,7 @@ export class Recurly extends Emitter {
     if (!this.configured) {
       this.configured = true;
       this.emit('configured');
-      if (this.event) this.event.send({ name: 'configured' });
+      this.report('configured');
```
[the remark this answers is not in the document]
chrissrogers: Mm I prefer the explicit reference, especially for calls. It's easier to know the intended reference. Dereferencing is nice for React when you're mixing values into long lines, but here it's much better to be explicit.

lib/recurly.js
chrissrogers: I don't think it would be very useful to separate it since it's only used once and isn't really a concern of the class, but just of this one routine.

lib/recurly.js
chrissrogers: While it's true that a const can have its properties mutated, I don't think they should be.

lib/recurly/pricing/subscription/index.js
chrissrogers: If the object returned is a reference to the currentPlan, it can be mutated in the caller and affect the original object used by the SubscriptionPricing instance.

lib/recurly/pricing/subscription/index.js
chrissrogers: common way of detecting if an element is in an Array

lib/recurly.js
gla7: Ah, so you assign it with `let` to let people know attributes will be added/deleted. Got it.

lib/recurly.js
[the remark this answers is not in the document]
gla7: Oh OK I see, yeah has potential to be confusing.

PR #471, opened 2018-07-27
Fix pricing calculation for multiple addons

rstrand:
Fixes a problem where only the first addon is calculated when the pricing module is attached to a form which already has multiple addon elements

### DIFF
--- a/lib/recurly/pricing/subscription/attachment.js
+++ b/lib/recurly/pricing/subscription/attachment.js
@@ -76,15 +76,15 @@ export default class Attachment extends Emitter {
 
     if (updateAddon) {
       pricing = pricing.then(() => {
-        elems.addon.forEach(elem => {
+        return Promise.all(elems.addon.map(elem => {
           const plan = this.pricing.items.plan;
           const code = dom.data(elem, 'recurlyAddon');
           const quantity = dom.value(elem);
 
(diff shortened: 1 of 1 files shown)


rstrand:
Rebased on the latest release. I'm not sure why the automated tests can't start the browsers but the phantomjs test passes.

chrissrogers:
This looks like a good solution at first glance. I'll have a run through with it shortly and let you know what I find.

As for the specs, Saucelabs will only run tests from our github organization due to being unable to decrypt the repo secrets.

chrissrogers:
To update you here, I haven't been able to reproduce the error case successfully. When I pre-populate a checkout form with a plan and add-ons with various quantities for the add-ons, the checkout pricing instance will pick up all of the add-ons and price accordingly. Do you have a specific repro scenario I can apply to see this occur?

rstrand:
Hi, I'm sorry I didn't see your previous message and just got the merge notification. Does that mean you manage to reproduce the bug?

chrissrogers:
@rstrand yes certainly. I'd been testing using the CheckoutPricing class, which did not exhibit the issue. I was able to reproduce this and found that your fix was an excellent solution. This will go out in the next version, release date TBA.

PR #484, opened 2018-10-16
Backfill AdyenHPP tests

adeitrick:
Adds missing tests for the Adyen functionality, specifically for payload validations after invoking the payment modal.


chrissrogers:
Looking good +1

PR #485, opened 2018-10-16
Fix - TypeError: Cannot read property 'brand' of undefined

gbakernet:
## Issue

We are seeing the 100k errors in production since Sept 27.
```
TypeError: Cannot read property 'brand' of undefined
1
File https://js.recurly.com/v4/recurly.js line 1 col 41317 in t.<anonymous>
2
File https://js.recurly.com/v4/recurly.js line 1 col 10701 in t.n.emit
3
File https://js.recurly.com/v4/recurly.js line 7 col 59595 in [anonymous]
4
File https://js.recurly.com/v4/recurly.js line 7 col 59477 in t.value
5
File https://js.recurly.com/v4/recurly.js line 7 col 58819 in t.value
```

## Change

The following change checks that `state` is truthy before accessing `brand`

### DIFF
--- a/lib/recurly.js
+++ b/lib/recurly.js
@@ -319,7 +319,7 @@ export class Recurly extends Emitter {
       this.on(`hostedField:${eventName}`, ({ type }) => {
         const state = this.hostedFields.state[type];
         let meta = pick(state, ['valid', 'empty']);
-        if (state.brand) meta.brand = state.brand;
+        if (state && state.brand) meta.brand = state.brand;
         this.report(`hosted-field:${eventName}`, meta);
       });
     });


chrissrogers:
Thanks for the PR @gbakernet -- mind giving me a quick scenario to reproduce the error?

chrissrogers:
The change looks good by the way. I'll merge it soon, but want to get some further details in case there is an underlying cause that we can track down.

gbakernet:
@chrissrogers It appears to happen when you focus on any hosted fields. You can see it in action in production here: 

1. https://elements.envato.com/subscribe (you'll be redirected to create an account maybe)
2. Select Credit Card payment
3. Focus on the Card Number field (or any hosted field). 
4. Check the developer console.

I'm not familiar with the specifics of how it's implemented. But could dig through if need be.

chrissrogers:
It appears the underlying cause is simultaneous operation of a couple recurly.js instances on your checkout. It appears three message buses are running simultaneously, and thus when the instances that do not have hosted fields attached to them receive the message, they are incapable of pulling state.

We are in the process of upgrading Recurly.js to isolate its message buses, but given the single lane available via postMessage, it is currently dispatching messages from the hosted fields to all instances on your checkout page.

If it is straightforward to only use the one instance on your checkout page, then I highly recommend that, as other unexpected behaviors may arise. To that end, I'll be updating the documentation to highlight the risks.

I think this is a good change that will resolve the behavior in this specific case, but the complete solution will come when instance buses are properly isolated. More work will go into this in the coming months.

gbakernet:
@chrissrogers Thanks very much for your analysis, we've reduced instances to one which has suppressed the error. Feel free to close the PR.

chrissrogers:
Glad to hear it! Closing.

PR #489, opened 2018-10-30
Fixes coupon resolving when switching subscription plans

chrissrogers:
- When switching plans on a `SubscriptionCheckout` instance,
  the constituent discounts must be checked for compatibility and
  removed accordingly
- Fixes #488

PR #490, opened 2018-10-31
Fixes tabbing specs

chrissrogers:
- Fixes spec assertions which failed outside of spec run context due to async behavior
- Fixes incorrect references


chrissrogers:
Specs will pass once #489 is merged into master

PR #493, opened 2018-11-07
Updates browser platforms in CI

chrissrogers:
User basis has shifted, thus we should shift our testing platforms accordingly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "node": ">=10.0.0"
   },
   "browserslist": [
-    "last 2 versions",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "node": ">=10.0.0"
   },
   "browserslist": [
-    "last 2 versions",
```
chrissrogers: This has no affect on CI, but it does reflect our support intention for browsers like Firefox and Chrome.

PR #504, opened 2019-01-17
Fixes detection of tabbing order on mobile browsers

### LINKED ISSUE #503: Tabbing with multiple hosted fields in mobile Safari
Related: #396 

On forms with multiple hosted fields (like the [advanced-pricing example](https://github.com/recurly/recurly-js-examples/blob/master/public/advanced-pricing/index.html), where this issue is reproducible) tabbing (while in the iOS simulator or with a paired or attached physical keyboard) and next/previous field navigation (the up/down arrows above the keyboard) don't seem to work for leaving a hosted field once you've focused in it. Instead of the focus changing to the next input (it doesn't matter if it's hosted or not), there doesn't appear to be any input focused.

chrissrogers:
- HostedField iframes were being added to the list of tabbable destinations
  due to presence of a tabIndex property
- Removes registered hosted field iframes from the list of available
  tabbing destinations to prevent this behavior
- Fixes #503


lib/recurly/hosted-fields.js
```diff
@@ -160,8 +160,14 @@ export class HostedFields extends Emitter {
...
 
+  /**
+   * Provides an Array of tabbable Elements in the document body,
+   * excluding those which correspond to hosted field iframes
+   * @return {Array} Elements
+   */
   tabbableItems () {
-    return tabbable(window.document.body);
+    const fieldFrames = this.fields.map(f => f.iframe);
+    return tabbable(window.document.body).filter(el => !~fieldFrames.indexOf(el));
```
aaron-suarez: I'm not sure how I feel about the use of `!~` for readability, it's somewhat obscure. But your comment at the top of the function explains what its doing so it's not too bad.

chrissrogers: It's a hold-over from IE11 not supporting `includes`, peppered throughout the codebase. I'm very excited to remove it in the future.

PR #514, opened 2019-02-16
Updates various packages to latest major versions

chrissrogers:
- Includes #513


aaron-suarez:
Are there any breaking changes in any of these packages?

chrissrogers:
Yeah I need to update the webpack and babel configuration to fix the ES transpiler.

PR #519, opened 2019-04-20
Updates packages to latest usable versions

chrissrogers:


package.json
```diff
@@ -23,7 +23,8 @@
     "lodash.chunk": "^4.2.0",
     "lodash.isempty": "^4.4.0",
     "lodash.pick": "^4.4.0",
-    "promise": "^8.0.2",
+    "promise": "^8.0.3",
+    "qs": "^6.7.0",
```
chrissrogers: qs belongs in build dependencies. This wasn't a big deal because our build environments install all packages, but it does belong up here.

PR #525, opened 2019-06-14
Emit PayPal cancel event on frame close

### LINKED ISSUE #520: Paypal frame close event
Hi there. Is there any event I can listen to, when a user manually closes the paypal dialog? Looked in paypal docs and it says that the `onCancel` callback should handle that, so I tried to listen for `recurly.PayPal` `cancel` events, but no success. Also tried `error` with the same results.
I been digging a bit into the code and found that the Frame is emitting `error` and `token`, but didn't find any `cancel` emit. But may be I missed something (not using braintree)
I am using version 4.9.6.
Thanks in advance

aaron-suarez:
Fixes #520 

This allows library users to listen for a `cancel` event that is emitted when the popup window for PayPal closes.

Example usage:
```javascript
this.paypal.on('cancel', () => {
    // react to the window closing
});
```

### DIFF
--- a/test/frame.test.js
+++ b/test/frame.test.js
@@ -28,12 +28,6 @@ describe('Recurly.Frame', function () {
       this.frame = this.recurly.Frame({ path });
       done();
     });
-
-    // Returns a spy so frame.bindWindowCloseListener does not cancel the tick
-    const windowStub = { close: () => { windowStub.closed = true; }, closed: false };
-    window.open = sinon.spy(() => windowStub);
-
-    sinon.stub(window.document.body, 'appendChild').callsFake(relay => relay.onload());
   });
 
   afterEach(function () {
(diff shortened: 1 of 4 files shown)


test/frame.test.js
```diff
@@ -15,7 +15,10 @@ describe('Recurly.Frame', function () {
       publicKey: 'test',
       api: `//${window.location.host}/api`
     });
-    window.open = sinon.spy();
+
+    // Returns a spy so frame.bindWindowCloseListener does not cancel the tick
+    window.open = sinon.spy(() => sinon.spy());
```
aaron-suarez: This is necessary because of this line of code https://github.com/recurly/recurly-js/blob/aac745c187f713bdfed4c739b10574f5471acd12/lib/recurly/frame.js#L103

test/frame.test.js
```diff
@@ -17,7 +17,8 @@ describe('Recurly.Frame', function () {
     });
 
     // Returns a spy so frame.bindWindowCloseListener does not cancel the tick
-    window.open = sinon.spy(() => sinon.spy());
+    const windowStub = { close: () => { windowStub.closed = true; }, closed: false };
```
chrissrogers: This gives `Frame.destroy` a `.close()` method to call, which is triggered when 'close' is emitted. It's circular and confusing, but it ensures the window actually is closed. Previously, the tests would exit early because the error would hit the ceiling, but the reporter would not bubble the failure properly.

chrissrogers:
This will require a docs update

PR #527, opened 2019-06-17
3D Secure 2.0

chrissrogers:
This is an omnibus 3D Secure 2.0 feature branch pull request!

**Summary**

- Add 3D Secure 2.0 features and PSD2 compliance to your checkout flow -- regardless of whether you currently tokenize credit cards with Recurly.js
- [See our blog post for an introduction](https://blog.recurly.com/preparing-to-comply-with-new-psd2-regulations)
- For more detail, [See our PSD2 documentation](https://docs.recurly.com/docs/revised-payment-services-directive-psd2)

**Integration guide in brief**

1. Load Recurly.js on your checkout, and configure it with your public key: See [Getting Started](https://dev.recurly.com/docs/recurly-js-getting-started) in our docs.
2. Identify a place in your checkout flow that you would like to inject the 3D Secure authentication flow dialogue. This will be at minimum a `250px x 400px` iframe that your customers will interact with in case 3DS authentication is required. We generally recommend this occur near your payment submission button, or in a modal.
2. Collect customer credit card information on your checkout however you currently do. Typically, this is through Recurly.js tokenization. Pass this to your backend server to use in the Recurly API.
3. Attempt to create a purchase or subscription via the Recurly API. If you receive a decline with the error code `three_d_secure_action_required`, then in that decline response, you will find a `three_d_secure_action_token_id`. Retrieve this token id value, and send it back to your frontend checkout flow.
4. Pass the `three_d_secure_action_token_id` to Recurly.js (in place of `'my-action-token-id'`):
```js
const container = document.querySelector('#my-three-d-secure-dialogue-location');

const risk = recurly.Risk();
const threeDSecure = risk.ThreeDSecure({ actionTokenId: 'my-action-token-id' });
threeDSecure.on('error', error => {
  // error handling
  // We recommend displaying a message to your customer informing them that
  // their authentication was not successful and that they should try a different
  // payment method.
});
threeDSecure.on('token', token => {
  // token.type => 'three_d_secure_action_result'
  // token.id
  // pass this token id to your backend to retry your initial API request
});
threeDSecure.attach(container);
```
5. Once your customer has authenticated their payment method, and you have received the `'token'` event, you will have a token result. This token is a `three_d_secure_action_result` Token, which is distinct from a credit card token. Pass this token id back to your backend.
6. Add the `three_d_secure_action_result_token_id` to your prior API request as part of the `billing_info` attributes. This will retry your operation with the data necessary to process the 3D Secure authentication. From here, you may proceed with decline and success handling as normal.

[integration-guide]: https://dev.recurly.com/page/recurly-3d-secure-2-integration-guide


karma.conf.js
```diff
@@ -40,7 +40,7 @@ var staticConfig = {
   client: {
     captureConsole: true,
     mocha: {
-      timeout : 800000, // 800 seconds
+      timeout: 3000, // 10 seconds
```
aaron-suarez: 3 seconds?

PR #530, opened 2019-06-20
ThreeDSecure reporting, UUID generation, and test suite cleanup

chrissrogers:
- Adds `ThreeDSecure` event reporting
- All UUIDs are now base64 encoded
- **Tests:** Adds tests to cover previously-uncovered sections of the lib: `Recurly.destroy`, `Recurly.bankAccount`, `Recurly.plan`, `Frame.relay`
- **Tests:** Cleans up frame tests to eliminate dangling requests occurring after the mock server has hung up.


lib/recurly/risk/three-d-secure/three-d-secure.js
```diff
@@ -66,11 +71,17 @@ export class ThreeDSecure extends RiskConcern {
...
+      this.strategy.attach(container);
+      this.report('attach');
+    });
   }
 
   remove () {
-    this.whenReady(() => this.strategy.remove());
+    this.whenReady(() => {
+      this.strategy.remove()
+      this.report('remove');
```
chrissrogers: I debated placing these reporters in `RiskConcern` but with the readiness need it made more sense to place it here. When we in the future move our other `Fraud` concerns over to the `RiskConcern` class, we can re-assess the SOC.

PR #531, opened 2019-06-21
Updates StripeStrategy to resolve only the PaymentIntent.id

chrissrogers:
- All we need is the `paymentIntent.id`, so this plucks only that value when digesting the `handleCardAction` result.
- **Bonus:** All UUIDs are now [nanoids](https://github.com/ai/nanoid#readme) -- better readability and more than sufficiently unique at 16 chars


lib/recurly/frame.js
```diff
@@ -46,7 +46,7 @@ class Frame extends Emitter {
   constructor ({ recurly, path, payload = {}, ...options }) {
     super();
     this.recurly = recurly;
-    this.id = `${this.recurly.id.split('-')[0]}-${uuid().split('-')[0]}`;
+    this.id = `${this.recurly.id}-${uid()}`;
```
chrissrogers: No need to parse these anymore since our nanoid charset does not include hyphens.

lib/recurly/risk/three-d-secure/strategy/stripe.js
```diff
@@ -44,7 +44,8 @@ export default class StripeStrategy extends ThreeDSecureStrategy {
         if (result.error) {
           return this.threeDSecure.error('3ds-auth-error', { cause: result.error });
         }
-        this.emit('done', result.paymentIntent);
+        const { id } = result.paymentIntent;
+        this.emit('done', { id });
```
chrissrogers: Here's the relevant update for plucking only the `id` from the `PaymentIntent` result

lib/util/uid.js
```diff
@@ -0,0 +1,12 @@
+import nanoid from 'nanoid/generate';
+
+const CHARSET = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
```
chrissrogers: I'm using the custom charset in order to remove hyphens and underscores. We use those in some url slugs, so this helps simplify parsing.

PR #532, opened 2019-06-27
Update adyen.js

pcote-stingray:
I've added the brandCode parameter, which allows to skip the payment method selection and save the user a click.
This is great if you need to support only one payment method using Adyen, in our case iDEAL.


pcote-stingray:
This needs to be supported at the recurly api level first...

PR #544, opened 2019-08-30
Add eslint

dbrudner:
Adding eslint and enforcing rules that are already established as convention within package scope. Also running eslint --fix and making changes to resolve linting errors.

At this point, this is more of a proposition than anything else. Hold off on merging pending further discussion. Thanks!


.eslintrc.js
```diff
@@ -0,0 +1,47 @@
...
+            'error',
+            2
+        ],
+        'linebreak-style': [
+            'error',
+            'unix'
+        ],
+        'quotes': [
+            'error',
+            'single'
```
chrissrogers: I think to avoid escapes within quotes we should set `{ "avoidEscape": true }`

Makefile
```diff
@@ -44,4 +44,10 @@ node_modules: package.json
 clean:
 	@rm -rf node_modules build
 
+lint:
+	./node_modules/.bin/eslint ./lib
+
+lint-fix:
+	./node_modules/.bin/eslint ./lib --fix
```
chrissrogers: Let's use `$(BIN)` here, add a `node_modules` dependent target, and move it above the `node_modules` target.

Issue #545, opened 2019-09-03
Fraud - Uncaught Error due to missing form


cfortuner:
Hi!

We are seeing following error when the form is removed from the page before Fraud is ready

```
code: "fraud-data-collector-missing-form"
message: "There was an error finding a form to inject the data collector fields using selector 'undefined'"
name: "fraud-data-collector-missing-form"
```

This is the error that is being thrown:
https://github.com/recurly/recurly-js/blob/e6ee4d326641b99c05155cb6c5405321a56b55b3/lib/recurly/fraud.js#L37

Is there a way that we can trap this error ? Other suggestions welcome!

Thanks
Colin

cfortuner:
Another error that seems to be related
![image](https://user-images.githubusercontent.com/5430709/64211463-17e8f680-ce5b-11e9-87d3-81e77f3c1577.png)

cfortuner:
I *think* what is happening is the user is navigating away from the current view and we are removing the form / window.recurly before the fraud.on('ready') callback starts/completes/finishes?

chrissrogers:
Hi! Just wanted to let you know that I'm aware of this and will have a more in-depth look as soon as I can. As always, thanks for all of the great detail you've provided.

chrissrogers:
Would you mind sharing the fraud configuration that you're setting?

cfortuner:
```
    recurlyInstance.configure({
      publicKey: recurlyPublicKey,
      fraud: {
        kount: { dataCollector: true }
      }
    });
   
    //  ... call recurly.fraud.on('ready', callback) 
  }
```

chrissrogers:
Awesome, thanks for that -- lines up with our assumptions. It's looking right now like there isn't a great workaround for catching this error in this scenario, since there isn't a clean entry-point to the async routine that ends up throwing the error. I suspect we'll have that confirmed shortly and will have a patch in for the next version, but more on that soon. Thanks again for the report!

chrissrogers:
This is closing with the merge of #546, which will allow you to listen to an `error` event emitted by `recurly` to receive this error instead of having it hit the ceiling when encountered. This will go out in the next version -- no exact release date yet, but I'll notify here once it's out. Cheers!

PR #546, opened 2019-09-06
adds error emission instead of throwing in Fraud

dbrudner:
Resolving https://github.com/recurly/recurly-js/issues/545

### DIFF
--- a/lib/recurly/fraud.js
+++ b/lib/recurly/fraud.js
@@ -34,7 +34,8 @@ export class Fraud extends Emitter {
     if (form) {
       return this._form = form;
     } else {
-      throw errors('fraud-data-collector-missing-form', { selectors });
+      const missingFormError = errors('fraud-data-collector-missing-form', { selectors });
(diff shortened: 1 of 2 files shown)


lib/recurly/fraud.js
```diff
@@ -83,8 +84,9 @@ export class Fraud extends Emitter {
       done: (error, response) => {
         debug("response from recurly data collector", error, response);
         if (error) {
-          throw errors('fraud-data-collector-request-failed', { error });
-        } else if (response.content) {
+          const requestFailedError = errors('fraud-data-collector-request-failed', { error });
+          this.emit('error', requestFailedError);
+        } else if (response.content && this.form) {
```
dbrudner: Because we're not throwing an error anymore, without checking for `this.form`, an unhandled exception would get thrown from this line https://github.com/recurly/recurly-js/blob/emit-fraud-error/lib/recurly/fraud.js#L103

PR #549, opened 2019-09-09
Fixes loading Braintree libraries

miscer:
When loading the Braintree libraries the code checks if `window.braintree.client` and `window.braintree.threeDSecure` exist, but not `window.braintree`. The first time the `loadBraintreeLibraries` method is called it throws an error: `TypeError: "window.braintree is undefined"`. This PR adds the additional check for `window.braintree` to avoid the error 🙂


chrissrogers:
Great catch! Merging this now to incorporate into pending version `v4.10.2`

PR #557, opened 2019-09-23
updates Frame to use enum instead of string types

dbrudner:
I'm thinking there's a better way to implement this that doesn't involve exporting/import `TYPE`, but I don't know what that would be.

With a static property as suggested in the ticket, I wasn't able to figure how to use that outside of the class when creating an instance of Frame (i.e., `const frame = new Frame({ type: Frame.TYPES.IFRAME });` or something like `Frame({ type: this.constructor.TYPES.IFRAME });` or `Frame({ type: this.TYPES.IFRAME });`

If I'm missing something, I'm open to changing anything.

### DIFF
--- a/lib/recurly/frame.js
+++ b/lib/recurly/frame.js
@@ -11,7 +11,7 @@ const debug = require('debug')('recurly:frame');
  * @property {Recurly} recurly
  * @property {String} path            API endpoint to load
  * @property {Object} payload         Message payload to deliver to endpoint
- * @property {TYPE} [type=0]          window or iframe
+ * @property {TYPE} [type=Frame.TYPES.WINDOW]          window or iframe
  * @property {HTMLElement} container  only valid for type='iframe' --
  *                                      attaches iframe to this container
  * @property {Number} [height=450]    frame height
@@ -32,23 +32,20 @@ export function factory (options) {
 }
 
 /**
- * @enum {Number}
- */
-export const TYPE = {
-  WINDOW: 0,
-  IFRAME: 1
-};
-
-/**
  * Issues an API request to a popup window.
  *
  * @param {FrameOptions} options
  */
-class Frame extends Emitter {
+export class Frame extends Emitter {
+  static TYPES = {
+    WINDOW: 0,
+    IFRAME: 1
+  };
+
   static DEFAULTS = {
     height: 535,
     width: 450,
-    type: TYPE.WINDOW
+    type: Frame.TYPES.WINDOW
   };
 
   constructor ({ recurly, path, payload = {}, ...options }) {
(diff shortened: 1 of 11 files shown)


lib/recurly/frame.js
```diff
@@ -31,6 +31,14 @@ export function factory (options) {
   return new Frame(options);
 }
 
+/**
+ * @enum {Number}
+ */
+export const TYPE = {
+  WINDOW: 0,
+  IFRAME: 1
+};
```
chrissrogers: I'm thinking we can put this onto a static property on `Frame`

```
class Frame {
  static TYPES = {
    WINDOW: 0,
    IFRAME: 1
  };
}
```

lib/recurly/frame.js
```diff
@@ -40,7 +48,7 @@ class Frame extends Emitter {
   static DEFAULTS = {
     height: 535,
     width: 450,
-    type: 'window'
+    type: TYPE.WINDOW
```
chrissrogers: We'd then use the reference `Frame.TYPES.WINDOW`

lib/recurly/frame.js
```diff
@@ -11,7 +11,7 @@ const debug = require('debug')('recurly:frame');
  * @property {Recurly} recurly
  * @property {String} path            API endpoint to load
  * @property {Object} payload         Message payload to deliver to endpoint
- * @property {String} [type='window'] one of 'window' or 'iframe'
+ * @property {TYPE} [type=0]          window or iframe
```
chrissrogers: nit: `[type=Frame.TYPES.WINDOW]`

lib/recurly/risk/three-d-secure/strategy/adyen.js
```diff
@@ -133,7 +134,7 @@ export default class AdyenStrategy extends ThreeDSecureStrategy {
       redirect_url: adyenRedirectParams.url,
       ...adyenRedirectParams.data
     };
-    this.frame = recurly.Frame({ type: 'iframe', path: '/three_d_secure/start', payload, container })
+    this.frame = recurly.Frame({ type: TYPE.IFRAME, path: '/three_d_secure/start', payload, container })
```
chrissrogers: Moving the enum will obviate the need to import `TYPES` separately

```suggestion
    this.frame = recurly.Frame({ type: Frame.TYPES.IFRAME, path: '/three_d_secure/start', payload, container })
```

dbrudner: This was my initial implementation, but I'm getting a reference error with this in my test: 

image

image


I tried a few other ways to access TYPES, but none of them worked. Any ideas?

chrissrogers: Oh right, we'd just need to import Frame into the `adyen.js` module

```js
import Frame from '../../../frame';
```

or reference it from the recurly instance in this routine

```js
this.frame = recurly.Frame({ type: recurly.Frame.TYPES.IFRAME, path: '/three_d_secure/start', payload, container });
```

PR #558, opened 2019-09-24
Browserstack

chrissrogers:
**Summary**
- Replaces Saucelabs CI runtime with Browserstack
- Real devices, improved stability

**Changes**
- Updates request failure message detail
- Small improvements to some test runtimes. PayPal `window.open` stubbing in particular
- Small changes to Travis CI config
- Updates CI targets: Adds Android 9, removes Android 5 and 6, removes iOS 11

**Follow-up**
- Add iOS 13 once it's stable on Browserstack. Devices are currently on beta branches that crash on launch.


.travis.yml
```diff
@@ -4,10 +4,10 @@ node_js:
   - 10
 cache:
   directories:
-  - node_modules
+    - node_modules
 script:
   - make lint
-  - travis_wait make test-ci
+  - make test-ci
```
chrissrogers: Big improvement here. `travis_wait` prevents the buffer from streaming into the log. Results are live-observable again with this change.

PR #568, opened 2019-11-15
Elements

chrissrogers:
## Introduction

`Elements` is a new API for displaying and controlling UI components of Recurly.js.

Primarily, Elements will be used in the way Hosted Fields are used today -- to display credit card inputs on your payment pages.

Elements also brings with it a more flexible API for interacting with objects that render to your page. So while their main use-case will be to render card inputs, in the future they will also be used to interact with other UI components, like 3-D Secure challenges, Apple Pay, PayPal flows, price information, and risk/fraud detection.

### The way things are now

Currently, Hosted Fields expose their interface through `recurly.configure`:

```html
<div data-recurly="card"></div>
```

```js
recurly.configure({
  publicKey: 'my-public-key',
  fields: {
    card: {
      inputType: 'mobileSelect',
      style: {
        fontColor: '#010101'
      }
    }
  }
});

// as the user changes values in a hosted field
recurly.on('change', state => {
  // I want the text to be green when they focus
  if (state.fields.card.focus) {
    recurly.configure({
      fields: {
        card: {
          style: {
            fontColor: 'green'
          }
        }
      }
    });
  } else {
    recurly.configure({
      fields: {
        card: {
          style: {
            fontColor: '#010101'
          }
        }
      }
    });
  }
});
```

This presents a number of challenges, as changes to the look-and-feel, or behavioral changes to your hosted fields had to be made through subsequent calls to `recurly.configure`. Because this pattern creates a 1:1 relationship between a `Recurly` instance and a Hosted Field group, It is not possible to instantiate multiple groups of hosted fields. Interactions with those fields have to proxy their behaviors and state information through the `Recurly` instance. Ultimately, this prevents direct control of Hosted Fields, and leads to an inconvenient API.

### How this will improve with Elements

Elements expose their APIs directly, so that control and state are much more granular. Here is what this looks like:

```html
<div class="my-card-element"></div>
```

```js
recurly.configure('my-public-key');

const elements = recurly.Elements();
const card = elements.CardElement({
  style: {
    inputType: 'mobileSelect',
    fontColor: '#010101'
  }
});
card.attach('.my-card-element');

card.on('focus', state => {
  card.configure({ style: { fontColor: 'green' } });
});

card.on('blur', state => {
  card.configure({ style: { fontColor: '#010101' } });
});
```

This is just one example of how direct control of an Element will help improve the semantics of working with Recurly.js. Let's get to the reference.

### `recurly.Elements`

Creates a new `Elements` controller. You will use this to create individual `Element` instances. Under the hood, this controller enforces rules about how `Element` instances may be grouped, handles communication between `Element` instances, and facilitates tokenization.

##### `elements.CardCvvElement()`

These methods create `Element` instances, respective of their names. See [Element](#element)

##### 'submit'

Emitted when a submit action has occurred in one of the `Element` instances within the `Elements` controller. Typically, this is when a user hits their `<enter>` key.

---

### `Element`

An `Element` controls an individual UI component. They are created using one of the methods available on a `recurly.Elements` instance.

**Arguments**
See [current style documentation](https://developers.recurly.com/pages/recurly-js.html#styling-card-fields).

##### `element.attach(parent)`

Attaches an `Element` to the DOM.

**Arguments**

| Argument | Type | Description  |
| ---------- | ----- | ------------ |
| parent | `HTMLElement` | This is where the `Element` will attach. It is appended as a child to the parent |

##### `element.configure(options)`

Configures an `Element`. All options passed to the constructor are accepted here.

**Arguments**

See [current style documentation](https://developers.recurly.com/pages/recurly-js.html#styling-card-fields).

##### `element.remove()`

If attached, removes an `Element` from the DOM. Otherwise, does nothing.

### Events

Listen to these events using `element.on(eventName, (...args) => {})`.

#### 'change'

Emitted when the `Element`'s state changes.

**Signature**

| Argument | Type | Description  |
| ---------- | ----- | ------------ |
| state | `Object` | Mixed hash of the `Element` state. This will be structured as relevant divisions of the current [`RecurlyState`](https://developers.recurly.com/pages/recurly-js.html#fieldsubmit) |

#### 'focus'

Emitted when a user focuses on the `Element`.

#### 'blur'

Emitted when a user removes focus from the `Element`.

#### 'submit'

Emitted when a user presses the `<enter>` key while using the `Element`.

#### 'attach'

Emitted when the `Element` has completed attaching to the DOM.

#### 'remove'

Emitted when the `Element` has been removed from the DOM.


lib/recurly.js
```diff
@@ -296,9 +334,10 @@ export class Recurly extends Emitter {
       this.emit('error', ...args);
     });
 
-    if (this.bus) this.bus.destroy();
-    this.bus = new Bus({ api: this.config.api });
-    this.bus.add(this);
+    if (!this.bus) {
```
dbrudner: Why the change here? To prevent an unnecessary `this.bus.destroy()` call?

chrissrogers: Coming back to this after chatting some in-person. This commit message goes into this change in detail: https://github.com/recurly/recurly-js/commit/8dacd36fb5a5ab5fb135b306374ea6cdb739494c

karma.conf.js
```diff
@@ -28,18 +28,21 @@ const staticConfig = {
...
+      base: 'VirtualBoxEdge',
+      keepAlive: true,
+      uuid: process.env.VIRTUALBOX_EDGE_UUID
+    },
     VirtualBoxIE11Win7: {
       base: 'VirtualBoxIE11',
       keepAlive: true,
-      vmName: 'IE11 - Win7',
-      uuid: '9c440145-fbea-49ea-a22f-4162bfaaaa2f'
+      uuid: process.env.VIRTUALBOX_IE11_UUID
```
chrissrogers: These definitely belong in ENV vars

lib/recurly.js
```diff
@@ -140,10 +154,27 @@ export class Recurly extends Emitter {
...
+    return storageSet({ key: 'deviceId', value: id });
+  }
+
   get sessionId () {
     return storageFetch({ scope: 'session', key: 'sessionId', otherwise: uid() });
   }
 
+  set sessionId (id) {
+    return storageSet({ scope: 'session', key: 'sessionId', value: id });
+  }
```
chrissrogers: These setters were added in order to allow fields to override their request identifiers. This also makes it easier to establish DNT patterns

lib/recurly.js
```diff
@@ -229,6 +260,12 @@ export class Recurly extends Emitter {
       this.emit('configured');
       this.report('configured');
     }
+
+    // Since this readyState indicates synchronous preparation, we may now
+    // emit ready
+    if (this.readyState === 3) {
+      this.emit('ready');
+    }
```
chrissrogers: This new action is to become the default result when using `Elements`. Eventually, we'll be able to do away with readyState tracking altogether.

lib/recurly/bus.js
```diff
@@ -45,14 +45,12 @@ export class Bus extends Emitter {
     this.debug('message event received', messageEvent.data);
     const message = Message.createFromMessageEvent(messageEvent, this.debug.bind(this));
     if (!this.shouldHandleMessage(message)) return;
-    this.send(message, null, { exclude: [messageEvent.srcElement] });
+    this.send(message, null, { exclude: [messageEvent.srcElement, messageEvent.source] });
```
chrissrogers: `srcElement` is IE-only, which is the sensitive runtime this was added to protect, but it is best to generically exclude the source from the recipient list

lib/recurly.js
```diff
@@ -274,20 +311,21 @@ export class Recurly extends Emitter {
...
   parent () {
-    let reset = false;
-
-    if (this.hostedFields) {
-      // Check integrity of hostedFields. If fields are dead or do not match chosen selectors, reset
-      if (this.readyState > 1 && !this.hostedFields.integrityCheck(this.config.fields)) reset = true;
-    }
+    // Check integrity of hostedFields. If fields are dead or
+    // do not match chosen selectors, reset
+    let reset = (
```
dbrudner: _nit_ `reset` to me sounds like a function. Maybe `shouldReset` would be more appropriate?

lib/recurly/bus.js
```diff
@@ -63,23 +61,41 @@ export class Bus extends Emitter {
     this.debug(`added recipient. Total: ${this.recipients.length}`, recipient);
   }
 
+  /**
+   * Remnoves a recipient from the bus
```
dbrudner: `Removes`

lib/recurly/bus.js
```diff
@@ -63,23 +61,41 @@ export class Bus extends Emitter {
...
+   *
+   * @public
+   * @param {Window|Emitter} recipient to remove
```
dbrudner: This jsdoc comment is super helpful for type defining `recipient`

lib/recurly/bus.js
```diff
@@ -63,23 +61,41 @@ export class Bus extends Emitter {
...
 
-    this.debug(`sending message to ${this.recipients.length} recipients`, message);
-    this.recipients.forEach(r => {
-      if (~exclude.indexOf(r)) return;
+    if (!message.event) {
+      this.debug('discarding message due to lack of event name', message);
+      return;
+    }
+
+    const recipients = this.recipients.filter(r => !~exclude.indexOf(r));
```
dbrudner: Just from looking at it, I'm not sure how this filter would be working. I would think you'd have to do some kind of deep equals for the filter to be able to work correctly. I could also be misunderstanding this.

lib/recurly/element/card-cvv.js
```diff
@@ -0,0 +1,10 @@
+import Element from './element';
+
+export function factory (options) {
+  return new CardCvvElement(Object.assign({}, options, { elements: this }));
```
dbrudner: I think object spread syntax might be more readable for situations like this, but that's definitely like just my opinion man and nit picky

`return new CardCvvElement({ ...options, elements: this })`

lib/recurly/element/element.js
```diff
@@ -0,0 +1,389 @@
...
+   */
+  get window () {
+    return this.iframe.contentWindow;
+  }
+
+  /**
+   * @private
+   *
+   * FIXME: contentWindow access is very slow.
+   *   - we may wish to use a sentinel to track attached state,
```
dbrudner: Any thoughts on making a ticket to pick this up at some point?

lib/recurly/element/element.js
```diff
@@ -0,0 +1,389 @@
...
+   * Updates state to reflect changes signaled by the Element iframe
+   *
+   * @private
+   * @param  {Object} body message body, containing new state
+   */
+  onStateChange (body) {
+    debug('state change', this.id, body);
+    let newState = { ...body };
+    delete newState.type;
+    if (JSON.stringify(this.state) === JSON.stringify(newState)) return;
```
dbrudner: Is there any value in abstracting out these deep equal object comparisons into a util function? Seems like it'd be an unnecessary abstraction, but figured I may as well throw it out there.

lib/recurly/element/card-cvv.js
chrissrogers: Sure, I prefer the spread syntax as well. `Object.assign` is much faster, but that hardly matters here.

lib/recurly/element/element.js
chrissrogers: Yeah that would be useful, certainly. It's only potentially perceivable in IE11 when making repeated configure calls

lib/recurly/token.js
```diff
@@ -64,20 +95,58 @@ export function token (options, done) {
...
+ * connection, request, or server error, and `response` is the
+ * recurly service response. The generated token is accessed
+ * at `response.token`.
+ *
+ * When using Elements
+ *
+ * @param {Object} customerData Billing properties
+ * @param {Elements} options.elements an Elements instance containing one tokenizable Element
+ * @param {HTMLFormElement} [options.form] whose children correspond to billing properties via their
+ *                                       'data-reurly' attributes
```
dbrudner: `data-recurly`

lib/recurly/token.js
```diff
@@ -64,20 +95,58 @@ export function token (options, done) {
...
+ *
+ * @param {Function} done callback
+ *
+ *
+ * When using Hosted Fields
+ *
+ * @param {HTMLFormElement} options An HTMLFormElement whose children correspond to billing
+ *                                  properties via their 'data-reurly' attributes
```
dbrudner: `data-recurly`

lib/recurly/element/element.js
chrissrogers: Mm since there's only two in here, it's okay as-is for now. But I'd totally welcome making a util for the op as a separate PR 💥

lib/util/deep-assign.js
```diff
@@ -23,6 +23,6 @@ export default function deepAssign (target, ...sources) {
   return deepAssign(target, ...sources);
 }
 
-function isObject (item) {
-  return item && typeof item === 'object' && !Array.isArray(item);
+function isAssignableObject (item) {
+  return item && typeof item === 'object' && !Array.isArray(item) && !(item instanceof Node);
```
dbrudner: I think checking for `item` and `typeof item === 'object'` is redundant?

chrissrogers: The object check is to ensure it isn't something like a `Boolean` or `String`, since we don't want to assign on those.

chrissrogers: hmm yes, the first op can be removed 👍

lib/recurly/bus.js
chrissrogers: recipients and exclude are both flat `Array`s of exact Object references, so it will look something like this.

```js
this.recipients = [instance, instance2, instance3];
const exclude = [instance2];
const recipients = this.recipients.filter(r => !~exclude.indexOf(r));
=> recipients = [instance, instance3]
```

dbrudner: That makes perfect sense

PR #577, opened 2020-01-30
Updates fraud detection endpoint consumer

chrissrogers:
- uses /js/v1/risk/info endpoint


lib/recurly/fraud.js
```diff
@@ -62,22 +61,45 @@ export class Fraud extends Emitter {
...
-          this.emit('ready');
+          return this.emit('error', errors('fraud-data-collector-request-failed', { error }));
         }
+
+        const { profiles } = response;
+
+        profiles.forEach(profile => {
+          const { processor, params } = profile;
+          if (processor === 'kount') {
+            const configuredForm = this.recurly.config.fraud.kount.form;
```
snodgrass23: Do we still need the fallback here in case a form isn't specified in the fraud config?

lib/recurly/fraud.js
```diff
@@ -62,22 +61,45 @@ export class Fraud extends Emitter {
...
+            const scriptElement = document.createElement('script');
+            const initializerElement = document.createElement('div');
+            const sessionIdInputElement = dom.createHiddenInput({
+              'data-recurly': 'fraud_session_id',
+              'value': params.session_id
+            });
+            const initialize = () => {
+              // eslint-disable-next-line no-undef
+              const client = new ka.ClientSDK();
```
snodgrass23: should we log an error if `ka` isn't defined?

lib/recurly/fraud.js
```diff
@@ -62,22 +61,45 @@ export class Fraud extends Emitter {
...
-            this.attachedNodes.push(node);
-            form.appendChild(node);
-          });
-          this.emit('ready');
+          return this.emit('error', errors('fraud-data-collector-request-failed', { error }));
         }
+
+        const { profiles } = response;
+
+        profiles.forEach(profile => {
```
snodgrass23: I like the flexibility this offers in case we need to add more/updated processors. I would probably break each processor function out but that's something we can do when needed for the next one.

jake-recurly:
QA+1 - ran tests with Kount enabled. and saw the following from kount : {"912468":"Scorecard:  Billing Country not equal to BIN Country (Visa/MC) (1)","912472":"Scorecard: Device Data Collector Missing & Score > 35 (1)"}.  Indicating we did do the device check.   Rogers also mentioned I did not need to do the additional paypal, 3ds tests.

lib/recurly/fraud.js
chrissrogers: Yep exactly. Each profile can be thought of as an RPC instruction. If we want to add additional instructions, we can namespace the processor i.e. `{ processor: 'kount.instruction2' }`

PR #580, opened 2020-02-06
Fixes fraud form detection fallback

chrissrogers:
- better errors and handling for fraud form detection
- refractors Recurly.giftCard
- improves tests and coverage


test/support/fixtures.js
```diff
@@ -177,6 +177,8 @@ const iframe = opts => `
 
 const threeDSecure = opts => `<div id="three-d-secure-container"></div>`;
 
+const emptyForm = opts => `<form id="test-form"></form>`;
```
dbrudner: It looks like `opts` is unused here

lib/recurly/pricing/subscription/index.js
```diff
@@ -162,19 +162,19 @@ export default class SubscriptionPricing extends Pricing {
...
     var self = this;
 
     return new PricingPromise(function (resolve, reject) {
       unsetGiftcard();
 
-      if (!giftcardCode) return resolve();
+      if (!code) return resolve();
 
-      self.recurly.giftcard({ giftcard: giftcardCode }, function (err,gift_card){
+      self.recurly.giftcard({ code }, function (err,gift_card){
```
dbrudner: Maybe we can take this opportunity to clean this up?
```
self.recurly.giftcard({ code }, function (err, gift_card) {
```

lib/recurly/fraud.js
```diff
@@ -99,9 +99,17 @@ export class Fraud extends Emitter {
...
+            if (form) {
+              const nodes = [
+                sessionIdInputElement,
+                scriptElement,
+                initializerElement
+              ];
+
+              nodes.forEach(node => form.appendChild(node));
+              this.attachedNodes = nodes;
+            }
```
chrissrogers: This is the essential fix for the fraud form detection. If an impl lacked a `form` param or did not have a valid `hostedFields` instance, this would have errored on `appendChild`.

chrissrogers: This also assigns `attachedNodes` to restore the functionality of `destroy`.

lib/recurly/fraud.js
```diff
@@ -130,7 +138,7 @@ export class Fraud extends Emitter {
       return this._form = form;
     } else {
       const missingFormError = errors('fraud-data-collector-missing-form', { selectors });
-      return this.emit('error', missingFormError);
+      this.emit('error', missingFormError);
```
chrissrogers: I removed this return so that an error would return undefined instead. This was revealed in the course of adding tests.

PR #582, opened 2020-02-24
Correctly whitelist bin ranges

calebbarde:
This PR aims to correctly whitelist Discover brand cards (previously we had whitelisted cards under different brands running on the Discover BIN range).

This adheres to: https://www.discoverglobalnetwork.com/downloads/IPP_VAR_Compliance.pdf


test/validate.test.js
```diff
@@ -1,7 +1,7 @@
 import assert from 'assert';
 import { Recurly } from '../lib/recurly';
 
-describe('Recurly.validate', function () {
+describe.only('Recurly.validate', function () {
```
billkauf1: Remove the `.only`.

lib/recurly/validate.js
```diff
@@ -15,8 +15,25 @@ const debug = require('debug')('recurly:validate');
...
+          [60112000,60114999],
+          [60117400,60117499],
+          [60117700,60117999],
+          [60118600,60119999],
+          [64400000,65059999],
+          [65060100,65060999],
+          [65061100,65999999]
+        ];
+
+        return accepted_ranges.find(([minimum, maximum]) => firstEight >= minimum && firstEight <= maximum) !== undefined
```
dbrudner: I think there needs to be a semi-colon at the end of this line to pass the lint step in our build https://travis-ci.org/recurly/recurly-js/jobs/654252599?utm_medium=notification&utm_source=github_status

test/validate.test.js
calebbarde: but it runs faster this way tho :trollface:

dbrudner:
Looks like `Array.find` isn't supported in IE and is causing tests to fail https://travis-ci.org/recurly/recurly-js/jobs/654456792?utm_medium=notification&utm_source=github_status

I think we tend to shy away from polyfilling to minimize overhead. Maybe switching to use `indexOf` is the best solution?

calebbarde:
@dbrudner yup saw that - just hadn't had a chance to fix it yet :) agreed with not using a polyfill, i'll switch it to indexof

chrissrogers:
This is being introduced with #590

PR #584, opened 2020-03-10
Adds support for typescript declaration generation and testing

dbrudner:


dbrudner:
Just a note: this will need to be squashed before merging. I wanted to leave the commit history as is until then so changes to types and tests I've made are viewable.

Makefile
```diff
@@ -46,12 +47,21 @@ test-e2e-debug: build $(src) $(tests)
...
 lint: build
 	@$(eslint)
 lint-fix: build
 	@$(eslint) --fix
 
+tsc:
+	@$(tsc) -p types -outFile ./build/index.d.ts
+
+tsc-watch:
+	@$(tsc) -w -p types -outFile ./build/index.d.ts
```
chrissrogers: These new targets will need to be added to `.PHONY`

package.json
```diff
@@ -101,6 +101,8 @@
     "terser-webpack-plugin": "^2.3.2",
     "wdio-chromedriver-service": "^5.0.2",
     "webpack": "^4.41.5",
+    "typescript": "^3.8.3",
+    "uglifyjs-webpack-plugin": "^2.2.0",
```
chrissrogers: I think this was added inadvertently?

test/types/3d-secure.ts
```diff
@@ -0,0 +1,14 @@
...
+    });
+
+    threeDSecure.on('token', function() {});
+
+    threeDSecure.on('error', function() {});
+
+    let el: HTMLElement;
+
+    threeDSecure.attach(el);
+}
```
chrissrogers: Looks like all of the ts files are using a spacing width of 4 when the rest of the lib uses 2

.travis.yml
```diff
@@ -21,6 +21,10 @@ jobs:
       name: ESLint
       script: make lint
 
+    - stage: Typecheck
+      name: Tsc
```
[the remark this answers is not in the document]
chrissrogers: Should this run the type tests? I may be reading this incorrectly, but is it just running the compiler on the types?

Makefile
```diff
@@ -46,12 +47,21 @@ test-e2e-debug: build $(src) $(tests)
...
+	@rm -rf ./dist
 
 lint: build
 	@$(eslint)
 lint-fix: build
 	@$(eslint) --fix
 
+tsc:
+	@$(tsc) -p types -outFile ./build/index.d.ts
+
```
chrissrogers: nit: remove the linebreak here to group like targets.

types/recurly.ts
```diff
@@ -0,0 +1,34 @@
...
+    Risk: Risk;
+    token: Token;
+    validate: Validate;
+}
+
+declare global {
+    interface Window {
+        recurly: Recurly;
+    }
+}
```
chrissrogers: sorry to nitpick: some declarations have terminating semicolons and others do not. Should we lint types?

chrissrogers:
I'd just like to say that this is absolutely spectacular and I can't wait to merge it in.

.travis.yml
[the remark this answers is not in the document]
dbrudner: ^^Yup, that's correct. There's no traditional unit testing here, this is just generating the types first, then compiling the 'tests' against those generated types. If the types are incorrect, the build should break and indicate a change is needed.

types/recurly.ts
dbrudner: We definitely should, great suggestion.

The semi-colon stuff is pretty interesting. Apparently interfaces shouldn't have semi-colons at the end of their bodies, but types should https://github.com/typescript-eslint/typescript-eslint/issues/1444

There's no eslint rule to enforce semi-colons at the end of function bodies either.

I actually think the best route to enforce some consistency with semi-colon usage would be adding https://eslint.org/docs/2.0.0/rules/no-extra-semi

Let's have a conversation about it and figure out a plan 👍

.travis.yml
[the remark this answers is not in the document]
chrissrogers: How are the tests run? Looking at the `Makefile`, it appears the types are just compiled, but the tests aren't being run.

Makefile
chrissrogers: let's change the output to `./build/recurly.d.ts` to match the rest of the lib's output

test/types/bank-account.ts
```diff
@@ -0,0 +1,50 @@
+export default function bankAccount() {
+  const formEl = document.querySelector('form');
+
+  if (!formEl) return;
```
chrissrogers: Just noting on these early returns that the ts compiler is evaluating `formEl` as a _maybe-_`HTMLFormElement`, thus the early return escape hatch here will account for the code branching where `formEl` is `null`.

test/types/configure.ts
```diff
@@ -0,0 +1,41 @@
+export default function configure() {
+  window.recurly.configure('ewr1-BrfKUWEllwCxdpRZvZloaJ');
```
chrissrogers: s/ewr1-BrfKUWEllwCxdpRZvZloaJ/test-public-key

test/types/paypal.ts
```diff
@@ -0,0 +1,23 @@
+export default function paypal() {
+  const paypal = window.recurly.PayPal();
```
chrissrogers: Might add some testing for the `options` that can be passed in to the constructor here.

test/types/pricing/subscription.ts
```diff
@@ -0,0 +1,63 @@
+export default function checkoutPricing() {
```
chrissrogers: s/checkoutPricing/subscriptionPricing

test/types/pricing/subscription.ts
```diff
@@ -0,0 +1,63 @@
...
+    })
+    .catch(e => {
+      e.code;
+      e.classification;
+      e.message;
+      e.name;
+    })
+    .done(pricing => {
+      if (pricing) {
+        pricing.price.now.total;
```
chrissrogers: maybe add asserts for all of the price values here

types/dt-header
```diff
@@ -0,0 +1,5 @@
+// Type definitions for recurly-js 4.12
+// Project: https://github.com/recurly/recurly-js
+// Definitions by: Dave Brudner <https://github.com/dbrudner>
+//                 Chris Rogers <https://github.com/chrissrogers>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
```
chrissrogers: Let's generate this file in `create-types.js`

test/types/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "lib": ["es6", "dom"],
+    "noImplicitAny": false,
+    "noImplicitThis": false,
+    "strictNullChecks": false,
+    "strictFunctionTypes": false,
+    "esModuleInterop": true,
+    "types": []
```
dbrudner: This is necessary to prevent the typescript compiler from pulling in and testing/linting node_modules when running `dts`, which it would do by default if this wasn't specified and cause this error https://github.com/microsoft/dtslint/issues/270

chrissrogers:
🎉

PR #590, opened 2020-03-19
Adds union pay

lexac1:
Adds union pay as a card type.

### DIFF
--- a/lib/recurly/validate.js
+++ b/lib/recurly/validate.js
@@ -18,7 +18,7 @@ const TYPES = [
     pattern: {
       test: (n) => {
         let firstEight = parseInt(n.substr(0, 8), 10);
-        let accepted_ranges = [
+        const acceptedRanges = [
           [60110000,60110399],
           [60110500,60110999],
           [60112000,60114999],
@@ -31,7 +31,7 @@ const TYPES = [
         ];
 
         if (!(/^6[045]/.test(firstEight))) return false;
-        return accepted_ranges.indexOf(([minimum, maximum]) => firstEight >= minimum && firstEight <= maximum) !== undefined;
+        return validBinRange(acceptedRanges, firstEight);
       }
     },
     lengths: [16, 17, 18, 19]
@@ -41,7 +41,7 @@ const TYPES = [
     pattern: {
       test: (n) => {
         let firstEight = parseInt(n.substr(0, 8), 10);
-        let accepted_ranges = [
+        const acceptedRanges = [
           [62109400,62109499],
           [62212600,62292599],
           [62400000,62699999],
@@ -54,7 +54,7 @@ const TYPES = [
         ];
 
         if (!(/^6[2]/.test(firstEight)) && !(/^8[1]/.test(firstEight))) return false;
-        return accepted_ranges.indexOf(([minimum, maximum]) => firstEight >= minimum && firstEight <= maximum) !== undefined;
+        return validBinRange(acceptedRanges, firstEight);
       }
     },
     lengths: [16, 17, 18, 19]
@@ -93,6 +93,20 @@ const TYPES = [
   }
 ];
 
+/**
+ * Validates bin ranges for supported card types
+ *
+ * @private
+ */
+function validBinRange (acceptedRanges, firstEight) {
+  for (const binRange of acceptedRanges) {
+    let min = binRange[0];
+    let max = binRange[1];
+
+    if (firstEight <= max && firstEight >= min) return true;
+  }
+  return false;
+}
 /**
  * Validation error messages
  * @type {String}
@@ -160,20 +174,17 @@ export function cardNumber (number) {
 
 export function cardType (number, partial = false) {
   const str = parseCard(number);
-  console.log('parseCard(number):', parseCard(number));
-  
   const len = str.length;
   const card = find(TYPES, card => card.pattern.test(str) && (partial || ~card.lengths.indexOf(len)));
-  console.log('The card type:', card);
-  
+
   return card && card.type || 'unknown';
 }
 
 /**
  * Validates whether an expiry month is present or future.
  *
- * @param {Numer|String} month The 2 digit month
- * @param {Numer|String} year The 2 or 4 digit year
+ * @param {Number|String} month The 2 digit month
+ * @param {Number|String} year The 2 or 4 digit year
  * @return {Boolean}
  */
 
--- a/lib/recurly/validate.js
+++ b/lib/recurly/validate.js
@@ -99,11 +99,13 @@ const TYPES = [
  * @private
  */
 function validBinRange (acceptedRanges, firstEight) {
-  for (const [min, max] of acceptedRanges) {
+  for (let index = 0; index < acceptedRanges.length; index++) {
+    const [min, max] = acceptedRanges[index];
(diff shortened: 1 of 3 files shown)


lib/recurly/validate.js
```diff
@@ -119,8 +160,12 @@ export function cardNumber (number) {
 
 export function cardType (number, partial = false) {
   const str = parseCard(number);
+  console.log('parseCard(number):', parseCard(number));
+  
   const len = str.length;
   const card = find(TYPES, card => card.pattern.test(str) && (partial || ~card.lengths.indexOf(len)));
+  console.log('The card type:', card);
```
chrissrogers: Let's remove these `console.log` calls

lib/recurly/validate.js
```diff
@@ -15,8 +15,49 @@ const debug = require('debug')('recurly:validate');
 const TYPES = [
   {
     type: 'discover',
-    pattern: /^(3[0689]|6011|621094|62[24568]|64[4-9]|65|81[0-7])/,
-    lengths: [14, 15, 16, 17, 18, 19]
+    pattern: {
+      test: (n) => {
+        let firstEight = parseInt(n.substr(0, 8), 10);
+        let accepted_ranges = [
```
chrissrogers: nit: `acceptedRanges`

lib/recurly/validate.js
```diff
@@ -15,8 +15,49 @@ const debug = require('debug')('recurly:validate');
...
+          [60110000,60110399],
+          [60110500,60110999],
+          [60112000,60114999],
+          [60117400,60117499],
+          [60117700,60117999],
+          [60118600,60119999],
+          [64400000,65059999],
+          [65060100,65060999],
+          [65061100,65999999]
```
chrissrogers: nit: space after the commas

lib/recurly/validate.js
```diff
@@ -15,8 +15,49 @@ const debug = require('debug')('recurly:validate');
...
+        ];
+
+        if (!(/^6[045]/.test(firstEight))) return false;
+        return accepted_ranges.indexOf(([minimum, maximum]) => firstEight >= minimum && firstEight <= maximum) !== undefined;
```
chrissrogers: [`Array#indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) doesn't accept a function like this, and the return value will always be `-1` here, thus this comparison will always return true.

test/unit/validate.test.js
```diff
@@ -39,12 +39,18 @@ describe('Recurly.validate', function () {
...
+      assert(recurly.validate.cardType('650611009013942423') === 'discover');
+      assert(recurly.validate.cardType('601179999013942423') === 'discover');
+    });
+
+    it('should parse union_pay', function () {
+      assert(recurly.validate.cardType('6221261111113245') === 'union_pay');
+      assert(recurly.validate.cardType('6282000123842342') === 'union_pay');
+      assert(recurly.validate.cardType('6250941006528599') === 'union_pay');
+      assert(recurly.validate.cardType('8171999927660000') === 'union_pay');
+      assert(recurly.validate.cardType('8171999900000000021') === 'union_pay');
```
chrissrogers: Seems like these card numbers aren't properly testing the ranges. Let's add some that are out of range to test that they are not recognized as discover or union pay

lib/recurly/validate.js
chrissrogers: nit: these should be `const`s

test/unit/validate.test.js
lexac1: Sure thing. Those are test cards from the discover / union pay docs. I can add the out of range negative case checks.

lib/recurly/validate.js
lexac1: Refactored that one for both discover and union_pay.

test/unit/validate.test.js
```diff
@@ -39,12 +39,19 @@ describe('Recurly.validate', function () {
...
-      assert(recurly.validate.cardType('3050990900242198') === 'discover');
-      assert(recurly.validate.cardType('36003010814263') === 'discover');
-      assert(recurly.validate.cardType('6011111111111117') === 'discover');
-      assert(recurly.validate.cardType('6011000990139424') === 'discover');
-      assert(recurly.validate.cardType('6229252986142899') === 'discover');
-      assert(recurly.validate.cardType('8131917256371926354') === 'discover');
+      assert(recurly.validate.cardType('6010999990139424') !== 'discover');
+      assert(recurly.validate.cardType('6011040090139424') !== 'discover');
+      assert(recurly.validate.cardType('6011000090139424') === 'discover');
+      assert(recurly.validate.cardType('6011039990139424') === 'discover');
```
chrissrogers: nit: for these, you can use `assert.strictEqual`

```js
assert.strictEqual(recurly.validate.cardType('6011039990139424'), 'discover');
```

test/unit/validate.test.js
```diff
@@ -39,12 +39,19 @@ describe('Recurly.validate', function () {
...
 
     it('should parse discover', function () {
-      assert(recurly.validate.cardType('3050990900242198') === 'discover');
-      assert(recurly.validate.cardType('36003010814263') === 'discover');
-      assert(recurly.validate.cardType('6011111111111117') === 'discover');
-      assert(recurly.validate.cardType('6011000990139424') === 'discover');
-      assert(recurly.validate.cardType('6229252986142899') === 'discover');
-      assert(recurly.validate.cardType('8131917256371926354') === 'discover');
+      assert(recurly.validate.cardType('6010999990139424') !== 'discover');
+      assert(recurly.validate.cardType('6011040090139424') !== 'discover');
```
chrissrogers: What is the actual return value of this call? I'm thinking we should `assert.strictEqual` on the actual return instead of asserting inequality.

lexac1:
> Please add tests for validations on CUP/UPI partial card numbers.

Will do.

lib/recurly/validate.js
```diff
@@ -121,14 +176,15 @@ export function cardType (number, partial = false) {
...
   return card && card.type || 'unknown';
 }
 
 /**
  * Validates whether an expiry month is present or future.
  *
- * @param {Numer|String} month The 2 digit month
- * @param {Numer|String} year The 2 or 4 digit year
+ * @param {Number|String} month The 2 digit month
+ * @param {Number|String} year The 2 or 4 digit year
```
chrissrogers: 👍

lib/recurly/validate.js
```diff
@@ -52,6 +93,20 @@ const TYPES = [
   }
 ];
 
+/**
+ * Validates bin ranges for supported card types
+ *
+ * @private
+ */
+function validBinRange (acceptedRanges, firstEight) {
+  for (const binRange of acceptedRanges) {
```
chrissrogers: syntax nit: this can be `for (const [min, max] of acceptedRanges) {}`

chrissrogers:
Looks like the failure may be do to the babel transform applied in the `for...of` loop -- IE11 doesn't support `Symbol` (https://travis-ci.org/github/recurly/recurly-js/jobs/665028875#L2283). You can get around that by using `Array#forEach` or a for loop

PR #591, opened 2020-04-07
Adds exports for types

dbrudner:
We discovered when creating types for react-recurly that our current types for recurly-js aren't exporting types. While the current types are sufficient for declaring Recurly, if a consumer needs to import a specific type right now, they would be unable to do so.

This change removes the declaration generation process, which is unecessary, in favor of writing our own declaration file that declares Recurly and also exports types from the lib.

Example usage (see test/types/recurly-recurly-error.ts):

```
import { RecurlyError } from "@recurly/recurly-js"

function myRecurlyErrorHandler (error: RecurlyError) {
    ...
}
```


test/types/recurly-error.ts
```diff
@@ -0,0 +1,10 @@
+import { RecurlyError } from '../../types';
```
dbrudner, line 1: This test shows what I'm trying to achieve. With this change, we can pull `RecurlyError` and other types out the definition and use them as needed.

chrissrogers:
Thinking we should export all types instead of a subset, to ensure broad access -- if a lib using these types as a dep needs something, then it will be available to them.

Makefile
```diff
@@ -68,5 +62,5 @@ clean:
 	@rm -rf node_modules build
 
 .PHONY: server server-http
-.PHONY: test-ci test-unit test-unit-ci test-unit-cov-ci test-e2e test-e2e-ci test-types
```
chrissrogers, line 71: This needs to remain 💥

test/types/tsconfig.json
```diff
@@ -6,6 +6,10 @@
     "strictNullChecks": false,
     "strictFunctionTypes": false,
     "esModuleInterop": true,
-    "types": []
+    "types": [],
+    "baseUrl": "./",
+    "paths": {
+      "@recurly/recurly-js": ["./index.d.ts"]
```
chrissrogers, line 12: is this essential?

types/3d-secure.ts
```diff
@@ -14,4 +14,4 @@ type ThreeDSecure = (riskOptions: RiskOptions) => ThreeDSecureEmitter;
 
 type Risk = () => { ThreeDSecure: ThreeDSecure };
 
-export default Risk;
+export { Risk };
```
chrissrogers, line 17: nit: `export type Risk`

types/index.d.ts
```diff
@@ -0,0 +1,31 @@
...
+export * from './apple-pay';
+export * from './bank-account';
+export * from './configure';
+export * from './elements';
+export * from './error';
+export * from './gift-card';
+export * from './paypal';
+export * from './recurly';
+export * from './token';
+export * from './validate';
```
chrissrogers, line 31: This is awesome.

test/types/tslint.json
```diff
@@ -1,6 +0,0 @@
-{
```
dbrudner, line 1: We no longer need this as we're not violating any dtslint rules.

PR #597, opened 2020-04-22
Exports server from test-server

dbrudner:


dbrudner:
I think we can use https://jestjs.io/docs/en/cli.html#--forceexit instead and not have to export anything to close the server

PR #598, opened 2020-04-22
Updates dependencies

chrissrogers:


chrissrogers:
I had to back out the `nanoid` update due to a compat issue with IE11.

PR #608, opened 2020-06-02
Support Bacs on RJS

calebbarde:
This PR adds support for Bacs in Recurly JS.


lib/recurly/bank-account.js
```diff
@@ -92,6 +93,10 @@ function token (options, done) {
     inputs.type = 'iban_bank_account';
   }
 
+  if ('branch_code' in inputs) {
+    inputs.type = 'bacs_bank_account';
+  }
```
chrissrogers, line 98: For this, to avoid making the coercion too complex as we add additional bank account types, I think we should start having merchants specify the `scheme_type` in their `token` call.

So in this method, we can accept `scheme_type` and cast it to `inputs.type` or accept `type` and allow it to pass into `inputs` directly.

That way we can move away from the black box decision-making we're already kinda suffering from in casting token type intent.

calebbarde: to round this up we are going to do "type" directly and coerce it to what RJS wants

lib/recurly/bank-account.js
```diff
@@ -24,6 +24,7 @@ const FIELDS = [
   'account_number_confirmation',
   'iban',
   'routing_number',
+  'branch_code',
```
chrissrogers, line 27: 👍

test/e2e/recurly.test.js
```diff
@@ -6,53 +6,98 @@ const {
...
 
-    assert.strictEqual(url.substring(0, url.indexOf('#')), `${environment().api}/field.html`);
+      assert.strictEqual(err, null);
+      assertIsAToken(token);
+    });
   });
 
-  it('creates a token', async function () {
-    const iframe = await $(sel.iframe);
+  describe.only('Bacs bank account', async function () {
```
chrissrogers, line 62: Will need to remove this `only`

test/unit/bank-account.test.js
```diff
@@ -128,39 +128,77 @@ apiTest(function (requestMethod) {
...
-        recurly.bankAccount.bankInfo({ routingNumber: 'test-routing-number' }, (err, bankInfo) => {
-          assert.deepEqual(bankInfo, { bank_name: 'test-bank-name' });
-          assert(!err);
+      describe('when given a routingNumber', function () {
+        it('responds with bank info', function () {
+          const { recurly } = this;
+          recurly.bankAccount.bankInfo({ routingNumber: 'test-routing-number' }, (err, bankInfo) => {
+            assert.deepEqual(bankInfo, { bank_name: 'test-bank-name' });
+            assert(!err);
+          });
```
chrissrogers, line 201: Awesome unit tests 👍

test/e2e/recurly.test.js
```diff
@@ -6,53 +6,98 @@ const {
...
+    beforeEach(init({ fixture: 'bank-account-bacs' }));
```
chrissrogers, line 63: I think this fixture is missing

test/e2e/recurly.test.js
```diff
@@ -6,53 +6,98 @@ const {
...
-    await (await $(sel.lastName)).setValue('Rambo');
+    const sel = {
+      output: '[data-test=output]',
+      form: '[data-test=form]',
+      submit: '[data-test=submit]',
+      nameOnAccount: '[data-test="name-on-account"]',
+      iframe: '.recurly-hosted-field iframe',
+      accountNumber: 'input[placeholder="Account Number"]',
+      accountNumberConfirmation: 'input[placeholder="Account Number Confirmation"]',
+      branchCode: 'input[placeholder="Branch Code"]',
```
chrissrogers, line 73: Might be easier to rely on a `data-test="branch-code"` pattern.

test/e2e/recurly.test.js
calebbarde: jeez idk what is up with you and always making me remove my `only`s

PR #614, opened 2020-06-17
Adds Becs support

calebbarde:


lib/recurly/bank-account.js
```diff
@@ -81,6 +83,8 @@ function token (options, done) {
     throw errors('missing-callback');
   }
 
+  console.log('this is a table')
```
adeitrick, line 86: is it?

calebbarde: are you

test/types/bank-account.ts
```diff
@@ -79,6 +79,24 @@ export default function bankAccount() {
     }
   });
 
+  const minimalBecsBillingInfo = {
+    type: "becs",
```
dbrudner, line 83: If `type` right now is only either "becs" or "becs", we can update type on BillingInfo.type to be `"bacs" | "becs"` instead of an empty string to lock down that property to only those two values https://github.com/recurly/recurly-js/blob/master/types/lib/bank-account.d.ts#L11

calebbarde: bah I super agree with this in theory, I'm finding it not possible to do this without having to add extra type inference, if i do a 

```
export type BankAccountType = "becs" | "bacs";

export type BillingInfo = {
# other stuff here ...
  type?: BankAccountType
}
```

then in the implementation code, i cannot pass `type: "bacs"` but instead need to to do 

```
let billingInfoType: BankAccountType = "bacs";
```
in order to get that stuff working correctly, or else I get an error saying its a string, not the correct type :( 

My concern with this is more around backwards compatibility and existing merchant integrations - is there a way around this concern? I'd love to do this!

dbrudner: **tldr: Agreed, leave it as a string.** Full explanation below if you or anyone else who comes across this is interested.

Ah yeah you're totally right, I missed this. This is a common typescript issue and has to do with typescript's type inference. You might've figured this out, but for anyone else who comes across:

In the test, [when `minimalBecsBillingInfo` is declared and `type` is set to "becs"](https://github.com/recurly/recurly-js/blob/8a4b14012ddab3bd0b9e98550ce26770d10ceb7e/test/types/bank-account.ts#L82), the ts compiler infers that `type` is any string, then when the [object gets passed into `bankAccount.token`](https://github.com/recurly/recurly-js/blob/8a4b14012ddab3bd0b9e98550ce26770d10ceb7e/test/types/bank-account.ts#L90), the types mismatch. In the pic below, note that the type for addressBecsBillingInfo.type is 'string'.

image

The way to fix on the implementation/test side is to make sure you're passing the right type, which can't be "string", as that doesn't pass the type assertion. There's a few ways to do this. You discovered one by declaring the correct type for `type` first before passing it into the object literal.

You can also directly pass an object into the `.token` method call and avoid typescript's type inference

image

You can also cast `addressBecsBillingInfo.type` as a `const` to tell the typescript compiler that the value is and will always be 'becs'. Note that the type for `addressBecsBillingInfo` in this case is "becs", not "string." IMO, **this is the best pattern for end users to follow and I think the most common approach for this situation in the typescript community**.

image

Having said all this, you're right that this change could break existing implementations (would break builds, but wouldn't break runtime). Also, I think the root cause could be hard for some merchant's to identify and fix. I'm cool with leaving this as a string.

PR #615, opened 2020-06-18
Fixes BankAccount type test

dbrudner:
This fixes a gap in our testing for bankAccount.token. It looks like I originally wrote the same test twice for the first parameter as an HTMLFormElement and forgot to test for BillingInfo.

### DIFF
--- a/test/types/bank-account.ts
+++ b/test/types/bank-account.ts
@@ -13,7 +13,23 @@ export default function bankAccount() {
     }
   });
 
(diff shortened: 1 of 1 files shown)


chrissrogers:
This passed in CI but the webhook failed to update GitHub: https://travis-ci.org/github/recurly/recurly-js/builds/699762275?utm_source=github_status&utm_medium=notification

PR #616, opened 2020-06-19
Use more inclusive terms

aaron-suarez:
The term "whitelist" is not very inclusive. "Allowlist" is better for Recurly's culture of diversity and inclusion.


chrissrogers:
👏

PR #617, opened 2020-06-23
Adds GitHub Action to create a standard GitHub Release

chrissrogers:


.github/workflows/release-notes.yaml
```diff
@@ -0,0 +1,53 @@
+# TODO: Add published to CDN date to description when the release is marked prerelease: false
+name: Create GitHub Release
+on:
+  repository_dispatch:
+    types:
```
dbrudner, line 5: Dope

Thoughts on adding a make command for this like we did with react-recurly?

chrissrogers: I think it has to wait on the job being able to accept a version parameter from the webhook. The dispatch capability isn't able to specify a tag for the job to consume. So for now, this is just a stand-in.

PR #618, opened 2020-06-24
Fixes previous tabbing for Elements on non-mobile browsers

chrissrogers:
- Ensures the mobile tabbing sink is only appended when using a mobile browser


test/e2e/behavior.tabbing.test.js
```diff
@@ -0,0 +1,73 @@
...
+
+  describe('when using distinct card Hosted Fields', function () {
+    beforeEach(init({ fixture: 'hosted-fields-card-distinct' }));
+
+    it(...tabsThroughTheForm());
+  });
+});
+
+function tabsThroughTheForm () {
+  return [`tabs across the form`, async () => {
```
dbrudner, line 49: quotes instead of backticks here? Doesn't really matter tho.

test/e2e/behavior.tabbing.test.js
```diff
@@ -0,0 +1,126 @@
...
+    let destinationReached = false;
+    const firstInput = await $('[data-test=arbitrary-input-0]');
+    const lastInput = await $('[data-test=arbitrary-input-1]');
+
+    if (environmentIs(DEVICES.IOS)) {
+      // iOS requires that each iframe be interacted with before focus directives succeed
+      const frames = await browser.$$('iframe');
+      for (frame of frames) {
```
dbrudner, line 58: Should be `for (const frame of frames) {` I think?

test/e2e/support/helpers.js
```diff
@@ -97,3 +150,42 @@ function environment () {
...
+    browserVersion,
+    platformName
+  } = browser.capabilities;
+  const browsers = Object.values(BROWSERS);
+  const devices = Object.values(DEVICES);
+  const isIos = platformName === 'iOS';
+  const isAndroid = platformName === 'Android';
+  const isMobile = isIos || isAndroid;
+
+  for (condition of conditions) {
```
dbrudner, line 174: `for (const condition of conditions) {`

PR #625, opened 2020-07-21
Adds Element.focus method to types

dbrudner:
Also updates Element test to include `.configure` and `.remove` and corrects return types for IndividualElement methods.

### DIFF
--- a/types/lib/elements.d.ts
+++ b/types/lib/elements.d.ts
@@ -60,12 +60,14 @@ export interface CardElement extends Emitter<ElementEvent> {
   attach: Attach<CardElement>;
   remove: () => CardElement;
   configure: (options: CardElementOptions) => CardElement;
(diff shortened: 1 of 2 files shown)


types/lib/elements.d.ts
```diff
@@ -60,12 +60,14 @@ export interface CardElement extends Emitter<ElementEvent> {
   attach: Attach<CardElement>;
   remove: () => CardElement;
   configure: (options: CardElementOptions) => CardElement;
+  focus: () => CardElement;
 }
 
 export interface IndividualElement extends Emitter<ElementEvent> {
   attach: Attach<IndividualElement>;
-  remove: () => Element;
```
dbrudner, line 67: `Element` here was referring to https://developer.mozilla.org/en-US/docs/Web/API/Element

PR #630, opened 2020-07-26
Adds test for rendering and loading fonts

dbrudner:


wdio.conf.js
```diff
@@ -123,6 +123,21 @@ function services () {
   return ['chromedriver'];
 }
 
+function imageComparisonService() {
+  return [
+    'image-comparison',
+    {
+      baselineFolder: path.resolve(__dirname, './test/snapshots'),
```
dbrudner, line 130: need to tighten up these settings

PR #633, opened 2020-08-05
Enhances types for BankAccount

dbrudner:
The goal here is to update the types for BankAccount so they reflect the runtime as accurately as they can. These changes are intended to provide type checking around the shape of the object getting passed to `BankAccount.token` by breaking down `BillingInfo` into its base payment types (ACH, SEPA, BACS, BECS).

### DIFF
--- a/test/types/index.d.ts
+++ b/test/types/index.d.ts
@@ -1,1 +1,2 @@
+// Minimum TypeScript Version: 3.1
 export * from '../../types';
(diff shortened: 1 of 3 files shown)


test/types/index.d.ts
```diff
@@ -1 +1,2 @@
+// Minimum TypeScript Version: 3.1
```
dbrudner, line 1: Specifying this for dtslint to run the same version of TypeScript that happens on DefinitelyTyped (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/recurly__recurly-js/index.d.ts#L6)

PR #636, opened 2020-08-12
Adds e2e test for kount

dbrudner:


lib/recurly.js
```diff
@@ -335,7 +335,9 @@ export class Recurly extends Emitter {
       return;
     }
 
-    if (!this.fraud) this.fraud = new Fraud(this);
+    if (!this.fraud || !this.fraud.dataCollectorInitiated) {
```
dbrudner, line 338: If fraud options gets passed when recurly.js is first initialized, the recurly object contains a fraud object, but nothing has been attached and dataCollector hasn't been initialized. If `recurly.configure` gets called after it's already been called, this check will always return false, and fraud will never reconfigure and attach.

test/e2e/recurly.test.js
```diff
@@ -174,7 +174,34 @@ describe('Recurly.js', async () => {
...
 
       assert.strictEqual(err, null);
-      assertIsAToken(token, expectedType="becs_bank_account");
+    });
+  });
+
+  describe('fraud', async function () {
+    beforeEach(init());
+    it('attaches an input with fraud session ID', async () => {
+      await browser.executeAsync((done) => {
```
dbrudner, line 183: Can't use `configureRecurly` for this because I need to pass in the `formElement` from the browser context

test/e2e/recurly.test.js
```diff
@@ -174,7 +174,36 @@ describe('Recurly.js', async () => {
...
+      const script = await $('form script');
+      await script.waitForExist(2000);
+
+      const scriptSrc = await script.getAttribute('src');
+      const kaxsdc = await $('.kaxsdc');
+      const fraudSessionIdInput= await $('input[data-recurly="fraud_session_id"]');
+      const sessionId = await fraudSessionIdInput.getValue();
+
+      await assert.strictEqual(sessionId.length, 32);
+      await assert.strictEqual(scriptSrc, `https://tst.kaptcha.com/collect/sdk?m=740001&s=${sessionId}`);
```
dbrudner, line 205: Should we be asserting the merchant ID that gets passed in the params here as well?

PR #637, opened 2020-08-12
Fix dt checkout action

dbrudner:


test/e2e/display.test.js
```diff
@@ -29,7 +29,11 @@ if (BROWSER !== 'Electron') {
           await clickBody();
 
           const diff = await browser.checkFullPageScreen('elements/card-element-pacifico');
-          assert.strictEqual(diff, 0);
+          if (BROWSER.includes('BrowserStackIos')) {
```
chrissrogers, line 32: Recommend using `environmentIs` here to branch tests based on env.

PR #638, opened 2020-08-12
Fixes error reporting

dbrudner:
image


lib/recurly/errors.js
```diff
@@ -316,12 +316,9 @@ function recurlyErrorFactory (definition) {
     }
 
     if (this.reporter instanceof Reporter) {
-      let type = 'client';
-      if (this.classification) type += `:${this.classification}`;
```
dbrudner, line 320: `classification` was always undefined here and the reporter would always send `type: 'client'`, regardless of the error.

chrissrogers (approved):

PR #653, opened 2020-10-22
Support setup intents on 3ds2

glaubenstein:
Update stripe 3ds2 strategy to support setup intents in addition to payment intents.


lib/recurly/risk/three-d-secure/strategy/stripe.js
```diff
@@ -39,18 +39,20 @@ export default class StripeStrategy extends ThreeDSecureStrategy {
     super.attach(element);
 
     this.whenReady(() => {
-      this.stripe.handleCardAction(this.stripeClientSecret).then(result => {
+      const handleAction = this.stripeClientSecret.startsWith('pi') ? this.stripe.handleCardAction : this.stripe.confirmCardSetup;
```
chrissrogers, line 42: `startsWith` will not work with IE11, and won't be transpiled currently. You can use instead:

```js
this.stripeClientSecret.indexOf('pi') === 0
```

chrissrogers:
Android unit test runs are failing on Browserstack, but are passing locally via emulator.

PR #661, opened 2020-12-15
Venmo support

efeygelson:


lib/recurly/venmo/strategy/braintree.js
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Loads Braintree client and modules
+   *
+   * @todo semver client detection
+   */
+  load ({ form }) {
+    debug('loading Braintree libraries');
+    this.form = form;
+
+    const part = after(2, () => this.initialize());
```
jpgnotgif, line 37: Should `2` be represented as a constant?

jpgnotgif (approved):

PR #665, opened 2021-01-19
Allows a PricingPromise to be provided to the ApplePay constructor

chrissrogers:
- Coerces the underlying pricing instance from the PricingPromise


karma.conf.js
```diff
@@ -1,5 +1,5 @@
 const {
-  BROWSER,
+  BROWSER = 'Chrome',
```
chrissrogers, line 2: This just defaults unit test runs to Chrome

types/lib/apple-pay.d.ts
```diff
@@ -26,7 +26,7 @@ export type ApplePayConfig = {
    * If provided, will override `options.total` and provide the current total price on the CheckoutPricing instance
    * when the Apple Pay flow is initiated.
    */
-  pricing?: CheckoutPricingInstance;
+  pricing?: CheckoutPricingInstance | CheckoutPricingPromise;
```
chrissrogers, line 29: This method technically will accept a generic 'PricingInstance' or 'PricingPromise', but our intent is to focus on `CheckoutPricing` going forward.

lib/recurly/apple-pay.js
```diff
@@ -191,6 +186,36 @@ class ApplePay extends Emitter {
     });
   }
 
+  /**
+   * Configures the methodology for pricing
+   *
+   * If a Pricing or PricingPromise instance is available, it will be used. Otherwise, a total
+   * will be expected.
+   *
+   * @param  {Object} options
```
csampson, line 195: Would probably be good to document the properties here

lib/recurly/apple-pay.js
```diff
@@ -191,6 +186,36 @@ class ApplePay extends Emitter {
...
+    } else if ('total' in options) {
+      this.config.total = options.total;
+    } else {
+      throw this.error('apple-pay-config-missing', { opt: 'total' });
+    }
+
+    if (pricing) {
+      this.config.pricing = pricing;
+      pricing.on('change', () => this.onPricingChange());
+      if (pricing.hasPrice) this.onPricingChange();
```
csampson, line 215: Could use a short comment explaining what's going on here

csampson (changes requested):

lib/recurly/apple-pay.js
chrissrogers: They’re documented in the constructor, so just didn’t want to repeat myself here.

csampson (approved):

PR #669, opened 2021-03-20
Adds prepare-release script

chrissrogers:
- Supports automation for release preparation: version bumping and changelog generation
- No release automation will be added


chrissrogers:
the spec failures here are due to image regression fixture mismatch on Firefox and focus flakiness on Edge. I should be able to  get to fixing this by early next week.

douglasmiller (approved):
👍

PR #684, opened 2021-06-21
Add test to support cuit tax identifiers

cyberxander90:
Tax identifier types could be `cpf` or `cuit`.


efeygelson (approved):

jpgnotgif (commented):
Looks good. Please add a negative test for a type other than `cpf` or `cuit`

cyberxander90:
> Looks good. Please add a negative test for a type other than `cpf` or `cuit`

There is a negative test [here](https://github.com/recurly/recurly-js/blob/79e54a122f9087a309307061bfe2bf88ab36df21/test/unit/token.test.js#L574)

brittany-wood (approved):

PR #687, opened 2021-07-20
Add support to SagePay 3DS2

cyberxander90:


jpgnotgif (commented):
@cyberxander90 Will you be updating the other 3D Secure strategies to also emit a default event name?

cyberxander90:
> @cyberxander90 Will you be updating the other 3D Secure strategies to also emit a default event name?

This is required only for Sagepay because they ignore the `TermUrl` where the listening event name is set up.

chrissrogers (approved):

PR #705, opened 2022-01-25
Adds unit test suite partitioning for IE 11

chrissrogers:
- Adds Unit test suite partitioning and implements partitioning for Internet Explorer 11 in CI
- Alleviates IE11 unit test timeouts in CI
- Also bumps some dependencies for compat with ARM architectures


cyberxander90 (approved):

test/unit/support/helpers.js
```diff
@@ -26,6 +26,22 @@ export function initRecurly (recurly, opts) {
   return recurly;
 }
 
+let describePartCount = 0;
+export function describePart (...args) {
+  const { SUITE_PARTITION } = window.__env__;
+  const partition = mod => SUITE_PARTITION === 'A' ? mod : !mod;
```
cyberxander90, line 32: @chrissrogers Is the execution running the tests in the same order? If not, I think we could be missing some tests.

cyberxander90: Wondering if will worth to override the mocha `describe` function in one place?

chrissrogers:
TODO: Fix Electron e2e tests

test/unit/support/helpers.js
chrissrogers: I did explore that path for a little while, but dropped it since there isn't a very good built-in way to monkey patch mocha's methods in browser test contexts.

Order is consistent, yes.

chrissrogers: On the `describe` override as well, you can't control for only partitioning the outermost describe calls. If we took the approach of striping the skips like I have here, then inner describe blocks will stripe along with the outermost. for example:

```js
describe(
  describe(
    describe()
    describe()
  )
);
```

would compose to:

```js
describe.skip(
  describe(
    describe.skip()
    describe()
  )
);
```

which would skip the entire suite instead of partitioning.

chrissrogers:
I reworked this to override `Mocha#it` for partitioning. This evens out the workload and makes the striped partitions more straightforward to check.

cyberxander90 (approved):

douglasmiller (approved):

PR #714, opened 2022-03-10
remove redirectURL from WorldpayStrategy

ajkelso:
Remove redirectURL from WorldpayStrategy so that we can dynamically change development vs production URLS in RA.


efeygelson:
https://github.com/recurly/recurly-js/pull/717 was merged, which branched off of this PR/branch

PR #747, opened 2022-08-03
Add Brand Picker Element

douglaslise:


lib/recurly.js
```diff
@@ -244,7 +249,7 @@ export class Recurly extends Emitter {
       this.config.currency = options.currency;
     }
 
-    this.config.coBrands = options.coBrands || [];
+    this.config.coBrands = options.coBrands || this.config.coBrands || [];
```
douglaslise, line 252: This is needed because `configure` method can be called multiple times, with incremental configuration options, so the existing should remain untouched.

lib/recurly.js
```diff
@@ -93,6 +93,11 @@ const DEFAULTS = {
     card: {
       selector: '[data-recurly=card]',
       style: {}
+    },
+    brand: {
+      selector: '[data-recurly=brand]',
+      style: {},
+      inputType: 'select',
```
chrissrogers, line 100: If it's just Elements we're planning to support here, then we can exclude this configuration default and the change in `hosted-fields.js`

tilley-kyle (approved):

PR #752, opened 2022-09-16
Cartes Bancaires and Co-Branding support

douglaslise:
Cartes Bancaires and Co-Branding support


lib/recurly/validate.js
```diff
@@ -117,34 +117,59 @@ function buildCompareValue (start, length, terminator) {
...
-
-        return compareValue >= start && compareValue <= end;
-      });
+      return findInRange(group.ranges, cardNumber);
     }));
+
+    if (group) {
+      return { type, group };
+    }
+  }).filter(function (typeWithGroup) { return !!typeWithGroup; });
```
chrissrogers, line 144: This could be reduced from two iterations to one by changing the above `map` to a `filter` and updating the return conditional to:

```
return group ? { type, group } : false;
```

chrissrogers: 👍 sure thing. A reduce would be possible, but it’s no more efficient and less straightforward.

chrissrogers: 👍 sure thing. A reduce would be possible, but it’s no more efficient and less straightforward.

chrissrogers (approved):

PR #757, opened 2022-10-06
Adding challenge window size selections

tilley-kyle:
We are adding in the capability for a merchant to select what size they would like their instance of RJS to display the 3ds challenge iframe.

To test. Supply the `challengeWindowSize` field in the call to 3DS Risk. Your site should be set on sandbox and you can use a sandbox 3ds card to trigger a transaction.
Example:
```js
risk.ThreeDSecure({
  actionTokenId: myActionTokenId,
  challengeWindowSize: risk.ThreeDSecure.CHALLENGE_WINDOW_SIZE_01_250_X_400,
});
``````
or
```js
risk.ThreeDSecure({
  actionTokenId: myActionTokenId,
  challengeWindowSize: '01',
});
``````

### DIFF
--- a/lib/recurly/frame.js
+++ b/lib/recurly/frame.js
@@ -48,6 +48,14 @@ export class Frame extends Emitter {
     type: Frame.TYPES.WINDOW
   };
 
+  static DIMENSIONS = {
+    '01': {height: '400px', width: '250px'},
+    '02': {height: '400px', width: '390px'},
+    '03': {height: '600px', width: '500px'},
+    '04': {height: '400px', width: '600px'},
+    '05': {height: '100%', width: '100%'}
+  }
+
   constructor ({ recurly, path, payload = {}, ...options }) {
     super();
     this.recurly = recurly;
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: lib/recurly/frame.js [hunk 1]
```diff
@@ -232,6 +240,8 @@ export class Frame extends Emitter {
   createIFrame () {
     const { container, url } = this;
     const iframe = document.createElement('iframe');
+    const dimensions = this.constructor.DIMENSIONS;
+    const windowSize = this.recurly.config.challengeWindowSize;
```
douglaslise, line 245: We can add a simple validation for this:
```suggestion

    const selectedWindowSize = dimensions[windowSize];
    if (!selectedWindowSize) {
      const validWindowSizes = Object.keys(dimensions).join(', ');
      throw new Error(`Invalid challengeWindowSize. Expected any of ${validWindowSizes}, got ${windowSize}`);
    }

```
Also, the newly created `selectedWindowSize` variable could be used directly when setting the attributes in line 255.

douglaslise (approved):
Just a small suggestion, feel free to ignore if not apply. 
In general it's a good job 👏

lib/recurly/frame.js
tilley-kyle: I like it. More validation the better I'd say

chrissrogers (approved):
Awesome PR!

PR #761, opened 2022-10-17
add auto-tabbing to combined card field

smagdicatrecurly:
Add "behavior" as an option on element to allow the following configuration:

```
elements.CardElement({
inputType: ‘text’
  style: {
    // this does not impact autotab behavior
  },
  behavior: {
    autoTabOnComplete: true,
    expiryYear: {
      maxLength: 2
    }
  }
});
```

This will also emit an autotab event the merchant can listen to.


chrissrogers:
Closing to address this feature with another approach. Code is great!

PR #773, opened 2022-11-10
Update Minimum Apple Pay Version to v4

cbarton:
Requires the device to not support v4 (Mojave)

Highlights include:
- Adds ApplePay merchant configuration property [`supportedCountries`](https://developer.apple.com/documentation/apple_pay_on_the_web/applepaypaymentrequest/2928612-supportedcountries)
- Supports $0 totals (common use case is free trials)

### DIFF
--- a/test/unit/apple-pay.test.js
+++ b/test/unit/apple-pay.test.js
@@ -5,7 +5,6 @@ import merge from 'lodash.merge';
 import omit from 'lodash.omit';
 import Emitter from 'component-emitter';
 import Promise from 'promise';
-import { Recurly } from '../../lib/recurly';
 import { initRecurly, apiTest, nextTick } from './support/helpers';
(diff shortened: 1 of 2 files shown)


test/unit/apple-pay.test.js
```diff
@@ -150,6 +151,29 @@ function applePayTest (integrationType, requestMethod) {
         });
       });
 
+      describe('when Apple Pay version not supported', function () {
+        let originalMethod = ApplePaySessionStub.supportsVersion;
+        beforeEach(function () {
+          ApplePaySessionStub.supportsVersion = () => false;
```
chrissrogers, line 157: Here we can leverage the existing Sinon sandbox. Can remove the afterEach since the sandbox is torn down afterward.

```
this.sandbox.stub(ApplePaySessionStub, 'supportsVersion').returns(false);
```

chrissrogers (approved):

PR #787, opened 2023-01-21
Adyen Payment Endpoints

cyberxander90:
Support Payment Methods with Adyen.

### DIFF
--- a/lib/recurly/errors.js
+++ b/lib/recurly/errors.js
@@ -39,8 +39,22 @@ const GOOGLE_PAY_ERRORS = [
     classification: 'internal'
   },
 ];
+const PAYMENT_METHODS_ERRORS = [
+  {
+    code: 'payment-methods-config-missing',
+    message: c => `Missing Payment Method configuration option: '${c.opt}'`,
+    classification: 'merchant'
+  },
+  {
+    code: 'payment-methods-not-available',
+    message: 'Payment Methods are not available',
+    classification: 'environment'
+  },
+];
+
 const ERRORS = [
   ...GOOGLE_PAY_ERRORS,
+  ...PAYMENT_METHODS_ERRORS,
   {
     code: 'not-configured',
     message: 'Not configured. You must first call recurly.configure().',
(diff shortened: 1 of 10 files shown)


lib/recurly/payment-methods/gateways/base.js
```diff
@@ -0,0 +1,41 @@
+import Emitter from 'component-emitter';
+
+const notImplementedError = new Error('not implemented');
+
+class Base extends Emitter {
```
gilv93, line 5: Hmmm. I'm not fully understanding the need for this class instead of just extending from Emitter?

lib/recurly/payment-methods/payment-methods.js
```diff
@@ -0,0 +1,165 @@
...
+
+  async handleAction (paymentResponse) {
+    try {
+      this.gatewayStrategy.handleAction(paymentResponse);
+    } catch (err) {
+      this.error(err);
+    }
+  }
+
+  // private
```
gilv93, line 52: Do we have any other patterns in RJS that indicate a method is private? Or is this pretty standard?

lib/recurly/payment-methods/gateways/base.js
cyberxander90: This is just an abstract class to define what methods need to be implemented in order to add a new PaymentMethod gateway strategy.
For stripe, we just need to implement these methods.
We can rip off this class but will be really hard to know what needs to be done to support Stripe Payment Methods.

lib/recurly/payment-methods/payment-methods.js
cyberxander90: it [exists](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields), but we are not using it in rjs, I don't even think is supported by webpack because the feature was delivered recently.
Other common patterns are pretending `_` to the function names, but we don't use that pattern either, so I just put that comment for developer reference.

lib/recurly/errors.js
```diff
@@ -39,8 +39,22 @@ const GOOGLE_PAY_ERRORS = [
     classification: 'internal'
   },
 ];
+const PAYMENT_METHODS_ERRORS = [
```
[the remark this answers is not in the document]
cyberxander90: I agree, this generic name could be really confused because credit cards are payment methods and they are not here. Let's talk about it in eng hangout.

gilv93 (approved):

arzitney:
QA +1

PR #788, opened 2023-02-14
[feat][apple-pay] send hostname to server and specify applicationData

cbarton:
Upgrades the native ApplePay integration to:
- Utilize the merchant Recurly subdomain as the `applicationData` if present
- Submit the customer browser hostname when retrieving the merchant config and starting the AP session.

Ensure that both the native and Braintree AP integrations continue to function.


chrissrogers (approved):

chrissrogers:
Only the publish phase failed on the specs. That failure should be addressed in a subsequent PR.

PR #794, opened 2023-03-24
feat(Apple Pay): validate and support more ApplePayPaymentRequest features

cbarton:
When Recurly supports more of the supported networks, ensure that the client will only query those that are supported by their browser to avoid any potential errors.

This also adds support to pass in an `ApplePayLineItem` into the `options.total` since the current options limit just the properties of the total line item to just the `label` and `total`. Note that using this interface will ignore `options.label`.

`options.label` is deprecated and optional in favor of the above. If not supplied with a dollar amount in `options.total`, the new `options.i18n.totalLineItemLabel` will be used, which is defaulted to "Total".

Added support for passing in an array of `ApplePayLineItem`'s via `option.lineItems`. This and the `total` are still overridden by the presence `options.pricing`.

Removes the payment request properties from the `ApplePay#config` property in favor of storing the full request in the psuedo-private `_applePayPaymentRequest` property.

Adds the boolean option `enforceVersion`, which will raise an `apple-pay-not-supported` error if the client does not support the minimum version for certain features, like `requiredShippingContactFields`.

Adds the ability to pass through all `ApplePayPaymentRequest` parameters to the session.


chrissrogers:
PR build shows as canceled, but all jobs therein passed

chrissrogers (approved):

PR #798, opened 2023-03-29
feat(apple pay): update tax when billing/shipping contact changes

cbarton:
When `options.pricing` is supplied to `recurly.ApplePay`, tax will be recalculated and stored based on the client's selection on the payment sheet for both billing and shipping contacts (shipping requires collecting the `postalAddress`).

Internally this changes the `SubscriptionPricing#items` schema from `items.shipping_address` to `items.shippingAddress` to match the `CheckoutPricing#items` schema. It does not change the `set.shipping_address` event that is emitted, however.

### DIFF
--- a/lib/recurly/apple-pay/apple-pay.js
+++ b/lib/recurly/apple-pay/apple-pay.js
@@ -186,12 +186,17 @@ export class ApplePay extends Emitter {
    */
   begin () {
     if (this.initError) return this.error('apple-pay-init-error', { err: this.initError });
+    if (this.config.pricing) {
+      const { address, shippingAddress } = this.config.pricing.items;
+      this._savedPricingState = { address, shippingAddress };
+    }
+
     delete this._session;
     this.session.begin();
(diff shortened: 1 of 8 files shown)


lib/recurly/apple-pay/apple-pay.js
```diff
@@ -377,3 +392,22 @@ export class ApplePay extends Emitter {
...
+    reprice = true;
+  }
+
+  if (pricing.items.shippingAddress != state.shippingAddress) {
+    pricing.items.shippingAddress = state.shippingAddress;
+    reprice = true;
+  }
+
+  if (reprice) pricing.reprice(done);
+  done();
```
chrissrogers, line 412: `done` can be called twice here if reprice is needed. Probably just need an `else`.

cbarton: updated

chrissrogers (approved):

PR #804, opened 2023-03-31
feat(apple pay): add support for `recurringPaymentRequest`, move to `options.paymentRequest`

cbarton:
Cleans up the interface for calling into `recurly.ApplePay` so that the
properties that drift down to the resulting payment request are not
intermixed with the SDK specific parameters.

As a result, deprecates `options.requiredShippingContactFields` in favor
of reading it off of the `paymentRequest.requiredShippingContactFields`.

Adds `options.recurring` that, when provided in combination with
`options.total`, will create a total that will include a monthly
recurring charge on the payment sheet.

`options.paymentRequest` will be treated as the source of truth, except
for when `options.pricing` is provided. This maintains the stance that
Pricing controls the `total` and `lineItems`.

Can be will be generated if the `total` line item is set to `paymentTiming='recurring'`:
```typescript
const applePay = recurly.ApplePay({
  total: '29.00',
  recurring: true,
});
```

OR specified as the `recurringPaymentRequest` option:

```typescript
const total: ApplePayLineItem = {
  label: 'My Subscription',
  paymentTiming: 'recurring',
  amount: '29.00',
  recurringPaymentIntervalUnit: 'month',
  recurringPaymentIntervalCount: 1,
  recurringPaymentStartDate: new Date(),
};

const applePay = recurly.ApplePay({
  paymentRequest: {
    total,
    recurringPaymentRequest: {
      paymentDescription: 'A recurring subscription',
      regularBilling: total,
      billingAgreement: 'Will recur forever',
    }
  }
});
```

### DIFF
--- a/lib/recurly/apple-pay/apple-pay.js
+++ b/lib/recurly/apple-pay/apple-pay.js
@@ -29,8 +29,10 @@ const I18N = {
  * @param {Recurly} options.recurly
  * @param {String} options.country
  * @param {String} options.currency
- * @param {String|Object} [options.total] either in dollar format, '1.00', or an ApplePayLineItem object that represents the total for the payment. Optional and discarded if 'pricing' is supplied
- * @param {String} [options.label] The short, localized description of the total charge. Deprecated, use 'i18n.totalLineItemLabel' if not using an ApplePayLineItem as the total
+ * @param {String|Object} [options.total] total for the payment in dollar format, eg: '1.00'. Optional and discarded if 'pricing' is supplied
+ * @param {String} [options.label] The short, localized description of the total charge. Deprecated, use 'i18n.totalLineItemLabel'
+ * @param {Boolean} [options.recurring] whether or not the total line item is recurring
+ * @param {Object} [options.paymentRequest] an ApplePayPaymentRequest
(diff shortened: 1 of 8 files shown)


chrissrogers:
The tests here just need an update to comply with the changeset in #802

lib/recurly/apple-pay/apple-pay.js
```diff
@@ -127,30 +133,23 @@ export class ApplePay extends Emitter {
...
-
-    if (options.form) {
-      this.config.form = options.form;
-      delete options.form;
-    }
+    if (options.recurly) this.recurly = options.recurly;
+    else return this.initError = this.error('apple-pay-factory-only');
+
+    if (options.form) this.config.form = options.form;
+    if (options.recurring) this.config.recurring = options.recurring;
```
chrissrogers, line 140: nit: this being a boolean will prevent it from setting through consistently when the option is provided. The side-effect of this will result in success, since `false` and `undefined` are functionally equivalent to the way this option is interpreted.

Simple fix would be to do `if ('recurring' in options) /* ... */`

chrissrogers (approved):

PR #807, opened 2023-04-04
add iframe titles on update

smagdicatrecurly:


lib/recurly/element/element.js
```diff
@@ -321,6 +321,11 @@ export default class Element extends Emitter {
     debug('update', id);
     this.container.className = classList;
     iframe.setAttribute('tabindex', tabIndex);
+    if (this.config.type === 'card') {
+      iframe.setAttribute('title', 'Billing information');
+    } else if (this.config.style?.placeholder?.content) {
+      iframe.setAttribute('title', this.config.style.placeholder.content);
+    }
```
chrissrogers, line 328: This condition should be reversed so that the `'Billing information'` title is only set when there is no placeholder content.

Thinking as well that we shouldn't have a default `title`, leaving it up to the integration author whether they want a `title` set by way of the placeholder attribute.

Another thing to consider is that the `CardElement` has its placeholder content shape of 
```
style.placeholder.content = {
  number: 'number placeholder',
  expiry: 'mm / yy',
  cvv: 'verification value'
}
```

which would render out into the title as `'[Object object]'` if referenced directly. We'll need a way to handle the title selection for that. I'd suggest reading `style.placeholder.content.number` if this is a card element and if that config value is present.

All in all, I'd suggest:

**lib/recurly/element/element.js**
```js
export default class Element extends Emitter {
  get title () {
    return this.config.style?.placeholder?.content;
  }

  // ...

  update () {
    // ...
    if (this.title) iframe.setAttribute('title', this.title);
    // ...
  }
}
```

**lib/recurly/element/card.js**
```js
export class CardElement extends Element {
  get title () {
    return this.config.style?.placeholder?.content?.number;
  }

  // ...
}
```

smagdicatrecurly: I think we are already on the same page here.

You mentioned:
"This condition should be reversed so that the 'Billing information' title is only set when there is no placeholder content. ...Another thing to consider is that the `CardElement` has its placeholder content shape of ...which would render out into the title as `'[Object object]'` if referenced directly. We'll need a way to handle the title selection for that. I'd suggest reading `style.placeholder.content.number` if this is a card element and if that config value is present."

The reason I have the following code is because of this reason you described above. Because CardElements placeholder content is an object
` ``
if (this.config.type === 'card') {
   iframe.setAttribute('title', 'Billing information');
```
this code would recognize the field or element as a card type and set that `CardElement` or `CardField`'s iframe title to "Billing Information". If it is NOT a CardElement and there is a placeholder config set for individual fields/elements it will use that placeholder to set it.

Also, we wouldn't have a default title for these iframes unless a placeholder is set or it is a card or hosted field element since we are setting that regardless.

Finally, I can change this to read from the card number placeholder if we have access to that in these files but it seems like billing information or credit card information could maybe work if we don't.

And thank you. I can separate some of the logic out a get title() function to separate some of the concerns.

Please let me know if I am misunderstanding any of your comment.

Comment (chrissrogers) [reply lib/recurly/element/element.js:328]: I think we're definitely close. The functionality in code now would evaluate to:

1. For a CardElement, use title 'Billing Information'
2. For any other Element type, use the `style.placeholder.content` option if it is set.

What I'm proposing is:

1. For a CardElement, use the `style.placeholder.content.number` option if it is set.
2. For any other Element type, use the `style.placeholder.content` option if it is set.
3. In any other situation, do not set a title attribute.

My comment above was a bit of stream of consciousness. I agree that your code wouldn't result in a CardElement displaying `[object Object]`, but that would be the result if, following what I started with at the beginning of my comment, the expression order was switched. Sorry, confusing.

Comment (smagdicatrecurly) [reply lib/recurly/element/element.js:328]: Awesome thanks Chris! Makes sense to me!

File: lib/recurly/element/element.js [hunk 2]
```diff
@@ -130,6 +130,7 @@ export default class Element extends Emitter {
     iframe.setAttribute('name', `${FRAME_NAME_PREFIX}${id}`);
     iframe.setAttribute('allowpaymentrequest', 'true');
     iframe.setAttribute('style', 'background: none; width: 100%; height: 100%;');
+    if (this.iframeTitle) iframe.setAttribute('title', this.iframeTitle);
```
jorge, line 133: Are single line conditionals without curly braces idiomatic JS?

jorge:
@chrissrogers we should add a PR template a la recurly-app to this repo

lib/recurly/element/element.js
chrissrogers: It matches the style of the rest of the project and passes linting, so it's okay.

jorge (approved):
🏅 L G T M 🏅

8eth:
QA+1 Done

PR #808, opened 2023-04-05
feat(Apple Pay): allow for customization of event updates

cbarton:
Adds the `options.callbacks` option which supports the following
callbacks:
- `onPaymentMethodSelected`
- `onShippingContactSelected`
- `onShippingMethodSelected`
- `onPaymentAuthorized`
to correspond with the `ApplePaySession` events.

A callback must return either the Apple Pay update object or a Promise
that resolves with that object.

That update object will be used to complete the `ApplePaySession` event
to open up the ability to fetch custom taxes on payment method selection,
update shipping costs on shipping contact/method selection and validate
email/phone addresses on authorization.

This does not remove the existing events being emitted but the workflow
should be updated to prefer using `onPaymentAuthorized` to complete the
purchase instead of async on the `token` event. The event that is sent
to that callback includes the `payment.recurlyToken` for use.

This also enhances the `token` event to include the raw payment event as
the second argument.

```javascript
const applePay = recurly.ApplePay({ 
  ..., 
  paymentRequest: { requiredShippingContactFields: ['email'] },
  callbacks: {
    onPaymentAuthorized: ({ payment: { recurlyToken, shippingContact: { emailAddress } } }) => {
      if (emailAddress === "test@example.com") return { errors: [new window.ApplePayError(...)] };
    },
  },
});
```

This also enhances the `token` event to include the raw payment event as the second argument.

### DIFF
--- a/lib/recurly/apple-pay/apple-pay.js
+++ b/lib/recurly/apple-pay/apple-pay.js
@@ -21,6 +21,17 @@ const I18N = {
   giftCardLineItemLabel: 'Gift card'
 };
 
+const CALLBACK_EVENT_MAP = {
+  onPaymentMethodSelected: 'paymentMethodSelected',
+  onShippingContactSelected: 'shippingContactSelected',
+  onShippingMethodSelected: 'shippingMethodSelected',
+  onPaymentAuthorized: 'paymentAuthorized'
+};
+
+const UPDATE_PROPERTIES = [
+  'newTotal', 'newLineItems', 'newRecurringPaymentRequest',
+];
+
 /**
  * Initializes an Apple Pay session.
  * Accepts all members of ApplePayPaymentRequest with the same name.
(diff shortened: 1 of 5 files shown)


cbarton:
Another option is to allow the client to pass in the various handlers on initialize and they call the `complete` function (that we control) to close out:

```typescript
const applePay = recurly.ApplePay({ 
  callbacks: {
    onPaymentMethodSelected: (event: ApplePayPaymentMethodEvent, 
                              complete: (update: ApplePayPaymentMethodUpdate) => void) => void;
  },
});
```

We would still support the emitters as well, but this would be the primary way of defining the explicit apple callbacks. 

The `token` event still works fine, though we should think about if we want the `paymentAuthorized` event to trigger after we receive the token (and emit it as part of the `event`) so the merchant can process the purchase to be more aligned with the UX of the payment sheet flow.

lib/recurly/apple-pay/apple-pay.js
```diff
@@ -383,6 +394,21 @@ export class ApplePay extends Emitter {
...
+    // listener accepts the `done` callback as an argument
+    if (!emitter.listeners(eventName).some(l => l.length >= 2)) {
+      emitter.emit(eventName, event);
+      handler(event);
+      return;
+    }
+
+    emitter.emit(eventName, event, (...args) => handler(event, ...args));
+  };
+}
```
chrissrogers, line 410: I think that this can become brittle when the integrator listens to the event more than once, where at least two of the listeners include a callback:

```
applePay.on('paymentAuthorized', ({ payment: { shippingContact } }, complete) => {
  if (shippingContact === "test@example.com") complete({ errors: [new window.ApplePayError(...)] });
  else complete();
});

applePay.on('paymentAuthorized', (event, complete) => {
  doSomethingAsync(event).then(() => complete());
});
```
This can introduce a race condition where the second listener would (a) be abandoned and (b) call the handler again, perhaps causing the ApplePay instance to produce an error?

There isn't a clean way to support multiple listeners with various return states, since it wouldn't be proper to collate or merge the return values...

So the alternative would be to prevent the integrator from listening with a callback more than once.

```
on (eventName, newListener) {
  if (this.listeners(eventName).some(l => l.length >= 2) && newListener.length >= 2) {
    // definitely open to suggestions on wording of this error.
    throw new Error(`The event '${eventName}' may only have one registered callback listener`);
  }

  return super.on(eventName, newListener);
}
```

cbarton: updated to support the explicit `options.callbacks`

chrissrogers (approved):

PR #815, opened 2023-04-20
fix(apple pay): simplify newRecurringPaymentRequest handling

cbarton:
[simplify newRecurringPaymentRequest handling](https://github.com/recurly/recurly-js/pull/815/commits/7ab71ce96ca58b6a754373ab4e618bd7f0273165) 
There are a variety of methods to initialize ApplePay with various
levels of integration with the native ApplePay request. On the simplest
level of configuration (`total`, `recurring` options), the merchant
might want to register the `onPaymentMethodSelected` callback to update
the total. Since the library creates the `recurringPaymentRequest` for
them automatically, we need to keep that total up to date if it changes.

We also manage the managementURL, so that property can be omitted when using the callback and added in automatically by the library.

[allow total to be specified on update callback](https://github.com/recurly/recurly-js/pull/815/commits/b0d5ea264193aca4c4eb958ba4b13e1f698d91e3) 
To simplify the integration for users using `options.total` and
optionally, `options.recurring`, along with `options.callbacks` to
update the total, the user can just specify the new `total` in the
callback and the library will convert that to `newTotal` for them to
hide the native Apple Pay internals.

[Default options.total to $0](https://github.com/recurly/recurly-js/pull/815/commits/2ee9e6aef788cdbbccdd64611e96b067ef8cae02) 
In 4.22, if `options.total` was provided as undefined, it would default
to $0 if no pricing was provided. This restores that behavior and goes
one step further by removing the requirement of `options.total`
altogether.

### DIFF
--- a/lib/recurly/apple-pay/apple-pay.js
+++ b/lib/recurly/apple-pay/apple-pay.js
@@ -278,6 +278,10 @@ export class ApplePay extends Emitter {
    * @private
    */
   completeSelection (onComplete, { errors, ...update } = {}) {
+    if (update?.total && !update.newTotal) {
+      update.newTotal = { ...this.totalLineItem, amount: update.total };
+    }
+
     if (update?.newRecurringPaymentRequest) {
       // Ensure that the newRecurringPaymentRequest has the required properties
       // from the recurringPaymentRequest stored from initialization.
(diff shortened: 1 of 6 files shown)


lib/recurly/apple-pay/apple-pay.js
```diff
@@ -350,6 +367,7 @@ export class ApplePay extends Emitter {
       data,
       done: (err, token) => {
         if (err) {
+          console.log(err);
```
chrissrogers, line 370: Should make this a `debug` call

chrissrogers (approved):

PR #822, opened 2023-04-28
Migrates CI from Travis to GitHub Actions

chrissrogers:


test/server/views/e2e/_head.html.ejs
```diff
@@ -9,6 +9,7 @@
       }
       .test-bed {
         background: white;
+        width: 330px;
```
chrissrogers, line 12: I locked down the width here to prevent small viewport drift from device to device causing large visual regression diffs. This is an issue particularly in Galaxy phones.

test/unit/apple-pay.test.js
```diff
@@ -77,22 +77,17 @@ const getBraintreeStub = () => ({
   },
 });
 
-describe('ApplePay', function () {
+const maybeDescribe = 'ApplePaySession' in window ? describe : describe.skip;
```
chrissrogers, line 80: Only need to test Apple Pay in browsers where customers would ever use it.

abachman (approved):

PR #827, opened 2023-05-12
Isolates public API test server into its own package

chrissrogers:


chrissrogers:
The integration tests will fail here until `react-recurly` is updated following this pull request with this manner of fixture server packaging.

jorge (approved):
@chrissrogers says the one error is okay

chrissrogers:
Right, error expected until recurly/react-recurly has updated its use of the fixture server on its main branch. [There's a branch here](https://github.com/recurly/react-recurly/tree/packages) for that.

PR #838, opened 2023-07-13
Accept Billing Address in AlternativePaymentMethods Submit

douglaslise:
Add option to pass a custom billingAddress for alternative payment methods that do not request it in their forms. 
Initially used for iDeal and Sofort.

Accepted billingAddress fields: `address1`, `address2`, `city`, `state`, `postalCode` and `country`.


gilv93 (approved):

czombo:
QA +1

PR #847, opened 2023-08-14
Prevent multiple 3ds instances

ajkelso:
When multiple purchases are attempted for the same customer/card within a similar timeframe, multiple 3DS challenges can occur which can cause issues with authorization on some gateways. This will prevent multiple challenges from existing and throw an error which indicates that another challenge already exists.


douglaslise (approved):

smagdicatrecurly:
QA+1

wsethbrown (approved):

PR #866, opened 2024-01-11
refactor rjs to resolve button reload error and add types

jsanderson1130:


types/lib/amazon-pay.d.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * The customer's locale. This is used to set the language rendered in the UI.
+   */
+  locale: string;
+
+  /**
+   * The currency of the payment.
+   */
+  currency: string;
```
douglasmiller, line 17: Shouldn't all of these options be optional like you have set `gatewayCode` to be?

types/lib/amazon-pay.d.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Specify which Payment Gateway in Recurly must handle the payment.
+   */
+  gatewayCode?: string
+};
+
+export type AmazonPayEvent = 'token' | 'error' | 'close';
```
douglasmiller, line 25: The `AmazonPay` class directly emits a `ready` event.

The `Frame`  class will emit `error`, `done`, and `close`. So you'll need to include `done` in this too.

types/lib/amazon-pay.d.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface AmazonPayInstance extends Emitter<AmazonPayEvent> {
+  /**
+   * Invokes the Amazon Payment Modal
+   */
+  start: (amazonPayOptions: AmazonPayOptions) => void;
+}
```
douglasmiller, line 32: The `AmazonPayInstance` also supports the `renderButton` and `attach` functions.

types/lib/amazon-pay.d.ts
douglasmiller: Actually, it looks like `currency` and `locale` are set based on the `region`. They should not be included as options.

douglasmiller: It also looks like there's a `sandbox` option that gets used in `renderButton`. I think that should be included as well as an optional option.

types/lib/amazon-pay.d.ts
jsanderson1130: Should I add `closed` instead of `close` as the `Frame` emits closed?

types/lib/amazon-pay.d.ts
```diff
@@ -0,0 +1,48 @@
...
+   * Specify which Payment Gateway in Recurly must handle the payment.
+   */
+  gatewayCode?: string
+
+  /**
+   * Sets button to Sandbox environment
+   */
+  sandbox?: boolean;
+
+  merchantId?: string;
```
douglasmiller, line 19: The `merchant_id` is fetched from the API before emitting the `ready` event. I don't think that it can be passed in as an option.

types/lib/amazon-pay.d.ts
```diff
@@ -0,0 +1,46 @@
+import { Emitter } from './emitter';
+
+export type AmazonPayOptions = {
+  /**
+   * 2 Digit Country Code
+   */
+  region: string;
```
douglasmiller, line 7: `region` should be optional as well.

types/lib/amazon-pay.d.ts
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Attaches an Element to the DOM, as a child of the specified parent target.
+   *
+   */
+  attach: (defaultEventName: string, gatewayCode?: string) => AmazonPayInstance;
+
+  /**
+   * Renders Amazon Pay button to the page
+   */
+  renderButton: (element: string | HTMLElement, amazonPayRenderButtonOptions: AmazonPayOptions & AmazonPayRenderButtonOptions) => AmazonPayInstance;
```
douglasmiller, line 43: I'm not seeing all of these additional parameters on the `renderButton` function. It looks to me like it only accepts a `element` parameter that is a `string`

types/lib/amazon-pay.d.ts
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Invokes the Amazon Payment Modal
+   */
+  start: (amazonPayOptions: AmazonPayOptions) => void;
+
+  /**
+   * Attaches an Element to the DOM, as a child of the specified parent target.
+   *
+   */
+  attach: (defaultEventName: string, gatewayCode?: string) => AmazonPayInstance;
```
douglasmiller, line 38: The `attach` function does not accept any parameters and does not return anything.

types/lib/amazon-pay.d.ts
douglasmiller: The `renderButton` does not return anything.

types/lib/amazon-pay.d.ts
```diff
@@ -0,0 +1,46 @@
...
+  buttonColor?: string;
+};
+
+export type AmazonPayEvent = 'ready' | 'token' | 'error' | 'close' | 'done';
+
+export interface AmazonPayInstance extends Emitter<AmazonPayEvent> {
+  /**
+   * Invokes the Amazon Payment Modal
+   */
+  start: (amazonPayOptions: AmazonPayOptions) => void;
```
douglasmiller, line 32: The `start` function does not accept any parameters.

douglasmiller (approved):

Issue #875, opened 2024-03-04
Recurly-JS is no longer updating the DefinitelyTyped repo


devinsimmons:
My team uses `react-recurly` and we want to upgrade to using amazon pay v2. However, `react-recurly` gets its type definitions from `DefinitelyTyped`. [You can see here that that repo lacks any type definitions for amazon pay](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/recurly__recurly-js/lib). While Recurly-JS has[ type definitions for amazon pay](https://github.com/recurly/recurly-js/blob/dbe1dbc80e5eef07de4ccd889057a5ace4eb144e/types/lib/amazon-pay.d.ts) and builds automated PRs for that repo, it's been sending PRs that do not pass `DefinitelyTyped` CI tests. A couple recent examples:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68799
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68636

We cannot update to amazon pay v2 until the new types are merged to that repo.

devinsimmons:
@chrissrogers it seems like at one point you may have been working on the DefinitelyTyped pull requests - I could always tidy up the most recent failed PR to address CI issues if someone from recurly reviews it

chrissrogers:
@devinsimmons I'd be happy to review those changes.

devinsimmons:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69067
made a PR in DT @chrissrogers

jakebailey:
Where does the integration bot live that produces the DT package? It's been failing for months, and the hand-written PR wasn't correct as it forgot to bump the package version. Fixing it should be straightforward...

chrissrogers:
@jakebailey The workflow is defined [here](https://github.com/recurly/recurly-js/blob/master/.github/workflows/publish-types.yaml)

jakebailey:
I don't think it is a fix in full; that PR does not include a `package.json` update to set the version, as that is no longer part of `index.d.ts`.

jakebailey:
I sent https://github.com/recurly/recurly-js/pull/886 which do the trick.

PR #886, opened 2024-07-11
Fix DT package.json version

jakebailey:
Per https://github.com/recurly/recurly-js/issues/875#issuecomment-2223172848

cc @douglasmiller


douglasmiller:
Thank you, @jakebailey. I hadn't noticed that DefinitelyTyped has a new requirement for a package.json file.

I've gone ahead and copied that file into this repo's types directory and updated the process that bumps the main package version to also update the types version in #887

jakebailey:
Sure, though you may end up in the situation where something in DT changes and your tooling overrides it. I guess that's not different than before where it was always updating the header too.

PR #887, opened 2024-07-11
Adding types/package.json

douglasmiller:
DefinitelyTyped now [requires that the types contain a package.json file](https://github.com/DefinitelyTyped/DefinitelyTyped?tab=readme-ov-file#packagejson).

This update will add that file and the process to keep it's version in sync with the main RJS package.

### DIFF
--- a/types/package.json
+++ b/types/package.json
@@ -1,0 +1,23 @@
+{
+    "private": true,
+    "name": "@types/recurly__recurly-js",
+    "version": "4.29.9999",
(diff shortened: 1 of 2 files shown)


types/package.json
```diff
@@ -0,0 +1,23 @@
+{
```
douglasmiller, line 1: Copied from DT repo

chrissrogers (approved):

PR #891, opened 2024-08-14
chore: extract loading braintree libraries to util/braintree

cbarton:
Consolidates loading Braintree dependencies into a shared module instead of copying it around each module that has a Braintree integration.

```js
import BraintreeLoader from 'path/to/./util/braintree-loader';

BraintreeLoader.loadModules('applePay', 'dataCollector')
  .catch(...)
  .then(...)
```


chrissrogers:
Excellent

PR #902, opened 2024-10-01
feat: Add support for tokenizing the CVV standalone

cbarton:
Allows `CardCvvElement` to tokenize the CVV by itself for use in CIT where the merchant wants the customer to confirm their CVV before checking out:

```js
const cvvElement = recurly.elements.CardCvvElement({});
cvvElement.attach(document.querySelector('#recurly-elements'));
```

### DIFF
--- a/packages/public-api-fixture-server/fixtures/field.html.ejs
+++ b/packages/public-api-fixture-server/fixtures/field.html.ejs
@@ -19,9 +19,13 @@
     }
 
     // Stub broker behavior
-    if (config().type === 'number') {
-      window.addEventListener('message', receivePostMessage, false);
+    function setStubTokenizationElementName (name) {
+      window.stubTokenizationElementName = name;
+      if (config().type === name) {
+        window.addEventListener('message', receivePostMessage, false);
+      }
     }
+    setStubTokenizationElementName('number');
 
     sendMessage(prefix + ':ready', { type: config().type });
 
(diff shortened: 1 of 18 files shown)


lib/recurly/element/cvv.js
```diff
@@ -0,0 +1,11 @@
+import Element from './element';
+
+export function factory (options) {
+  return new CvvElement({ ...options, elements: this });
+}
+
+export class CvvElement extends Element {
```
cbarton, line 7: Debating whether or not this calls for a new element class or we can add `supportsTokenization` to `CardCvvElement` to allow it to work stand alone.

packages/public-api-fixture-server/fixtures/field.html.ejs
```diff
@@ -19,7 +19,8 @@
     }
 
     // Stub broker behavior
-    if (config().type === 'number') {
+    var tokenizeElementType = config().recurly.fields.tokenizeElement || 'number';
```
chrissrogers, line 22: I'm thinking it would be best to just assign this behavior onto the window, since the security profile used in specs is permissive enough to do something like the following:

**field.html.ejs**
```js
window.stubTokenizationElementName = 'number';

function setStubTokenizationElementName (name) {
  window.stubTokenizationElementName = name;
}

// ...

function onToken (body) {
  // ...
  if (stubTokenizationElementName === 'cvv') {
    recurly.config.required = ['cvv'];
  }
  // ...
}
```

**test**
```js
element.iframe.window.setStubTokenizationElementName('cvv')
```

chrissrogers (approved):

PR #904, opened 2024-10-25
Allow billingAddress to be passed in via customer options on AlternativePaymentMethod

gilv93:
Adds alternative path for setting customer fields such as billingAddress when initializing AlternativePaymentMethod class

### DIFF
--- a/lib/recurly/alternative-payment-methods/gateways/adyen.js
+++ b/lib/recurly/alternative-payment-methods/gateways/adyen.js
@@ -1,4 +1,6 @@
 import Base from './base';
+import AlternativePaymentMethods from '../alternative-payment-methods';
+
 
 class AdyenGateway extends Base {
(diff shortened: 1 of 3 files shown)


lib/recurly/alternative-payment-methods/gateways/adyen.js
```diff
@@ -89,7 +96,7 @@ class AdyenGateway extends Base {
   }
 
   async submitWebComponent (billingAddress) {
-    this.customerBillingAddress = billingAddress;
+    if (this.customerBillingAddress === undefined) { this.customerBillingAddress = billingAddress; }
```
chrissrogers, line 99: Let's allow an override here if a new billingAddress is provided:

```js
if (billingAddress) {
  this.customerBillingAddress = billingAddress;
}
```

chrissrogers (approved):

PR #913, opened 2024-12-13
fix: added `payPalComplete` property to PayPal `DirectConfig` type de…

magicink:
`DirectConfig` appears to be missing the `payPalComplete` property mentioned in the documentation.

### DIFF
--- a/types/lib/paypal.d.ts
+++ b/types/lib/paypal.d.ts
@@ -1,13 +1,16 @@
 import { Emitter } from './emitter';
 
+export type PayPalDisplayConfig = {
+  locale?: string;
+  displayName?: string;
+  amount?: string;
+}
+
 export type BraintreeConfig = {
   braintree: {
     clientAuthorization: string;
   };
-  display?: {
-    locale?: string;
-    displayName?: string;
-    amount?: string;
+  display?: PayPalDisplayConfig & {
     enableShippingAddress?: boolean;
     shippingAddressOverride?: any;
     shippingAddressEditable?: boolean;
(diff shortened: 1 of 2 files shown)


chrissrogers:
Thanks for the contrib!

types/lib/paypal.d.ts
```diff
@@ -25,6 +25,7 @@ export type DirectConfig = {
     logoImageUrl?: string;
     headerImageUrl?: string;
   };
+  payPalComplete?: boolean;
 };
 
 export type PayPalConfig = BraintreeConfig | DirectConfig;
```
chrissrogers, line 31: I think to match and hint at the underlying logic, we might consider the following:

```
export type PayPalCompleteConfig = {
  payPalComplete?: boolean;
};

export type PayPalConfig = BraintreeConfig | DirectConfig | PayPalCompleteConfig;
```

magicink: I think this pull request should be closed because I'm confused about which version of the library this pertains to. By default, the documentation that initially displays (`latest`) shows this option under the PayPal section, but if you switch the documentation to `4.10.3` this is not an option. When I applied this to the version that is delivered via [js.recurly.com/v4/recurly.js](https://js.recurly.com/v4/recurly.js), this introduces strange behaviors with PayPal complete, but when left out, things work as expected. This leads me to believe the version served by js.recurly.com is actually `4.10.3` and not `latest`. So perhaps this actually isn't an option on the current (actual latest) version of the platform?

magicink: These changes have been applied. I created a new `PayPalCompleteConfig`. I took a best guess at what other properties would be needed.

> I think to match and hint at the underlying logic, we might consider the following:
> 
> ```
> export type PayPalCompleteConfig = {
>   payPalComplete?: boolean;
> };
> 
> export type PayPalConfig = BraintreeConfig | DirectConfig | PayPalCompleteConfig;
> ```

magicink:
@chrissrogers looks like the merge is being blocked by BrowserStack credential issues.

PR #915, opened 2025-02-27
Add fraud params to Apple-Pay token

jsanderson1130:
Update Apple-Pay token params to allow for `fraud` parameters.


Patrick-Duvall (approved):

lib/recurly/apple-pay/apple-pay.js
```diff
@@ -438,12 +438,24 @@ export class ApplePay extends Emitter {
...
       paymentMethod,
-      ...(this.config.form && normalize(this.config.form, NON_ADDRESS_FIELDS, { parseCard: false }).values),
+      ...combinedInputs.call(this),
       ...transformAddress(billingContact, { to: 'address', except: ['emailAddress'] }),
     };
   }
 }
 
+function combinedInputs () {
+  if (this.config.form) {
```
chrissrogers, line 448: I think it would be appropriate to parameterize `this` since this function isn't a member of the instance.

```
function combinedInputs(recurly) {
  if (recurly.config.form) {
  // ...
```

chrissrogers: You might want to do an early return for simplification purposes:

```
if (!this.config.form) {
  return;
}
```

PR #933, opened 2025-05-14
WIP

arzitney:


lib/recurly/token.js
```diff
@@ -203,7 +203,8 @@ function token (customerData, bus, done) {
       state
     } = inputs;
 
-    if (number && month && year) {
+    // QUESTION: Should I check for the presence of billing info too (billing_info_id && cvv)?
+    if ((number && month && year) || cvv) {
```
gilv93, line 207: yeah let's leverage the billing_info_id field being present

lib/recurly/risk/three-d-secure/strategy/cybersource.js
```diff
@@ -16,9 +16,11 @@ export default class CybersourceStrategy extends ThreeDSecureStrategy {
     }
 
     const { PREFLIGHT_TIMEOUT } = CybersourceStrategy;
+    // debugger
     const data = {
       gatewayType: CybersourceStrategy.strategyName,
       gatewayCode: gateway_code,
+      billingInfoId: recurly.config.billingInfoId,
```
gilv93, line 23: IIRC we don't expect this to be set in the `recurly.config` but something in the `options` param of the element (adjacent to these style options [here](https://recurly.com/developers/reference/recurly-js/?#reference-1))

PR #950, opened 2025-07-26
Enhances CVV validation by introducing card type-specific checks. Add…

Codestzx:
…s `cvvWithCardNumber` function to validate CVV length based on card type and updates tests to cover various card scenarios and edge cases.


Codestzx:
Sorry this was a mistake.

PR #987, opened 2026-06-26
feat(pricing): support sub-cent add-on prices via unit_amount_decimal

Jhonatanssal:
Fixed and flat add-on prices were displaying as -zsh.00 when priced below one cent. The Recurly API returns these prices in unit_amount_decimal (a string with up to 9 decimal places) while unit_amount is truncated to 0.)

-  Read unit_amount_decimal with unit_amount as fallback in the flat add-on path (calculations.js) and in tiered-pricing-calculator.js
-  Store per-add-on prices at 9-digit precision before decimalizeMember runs so the 2-place rounding pass does not truncate sub-cent values; decimalizeMember skips strings, so the stored value is preserved
-  Fix round() to correctly honor the digits parameter using the string-exponent trick; the previous implementation hardcoded e+2/e-2 making round(x, 9) always round to 2 decimal places
-  Add sub-cent-decimal-addon and sub-cent-tiered-decimal-addon fixtures with tests covering both unit_amount_decimal and unit_amount fallback code paths

### DIFF
--- a/lib/recurly/pricing/subscription/calculations.js
+++ b/lib/recurly/pricing/subscription/calculations.js
@@ -1,7 +1,7 @@
 import each from 'component-each';
 import find from 'component-find';
 import isEmpty from 'lodash.isempty';
-import { clampToZero, decimalizeMember, round } from '../../../util/decimalize';
+import decimalize, { clampToZero, decimalizeMember, round } from '../../../util/decimalize';
 import { getTieredPricingTotal, getTieredPricingUnitAmount, isTieredAddOn } from './tiered-pricing-calculator';
 
 /**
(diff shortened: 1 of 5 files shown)


Patrick-Duvall:
### ⚠️ `decimalize(_, 9)` changes the public output format for *all* add-ons, not just sub-cent ones

In `lib/recurly/pricing/subscription/calculations.js`:

```js
this.price.base.addons[addon.code] = decimalize(unitPrice, 9);
this.price.addons[addon.code]      = decimalize(totalPrice, 9); // DEPRECATED
```

`decimalize()` returns a **string** (`round(n, 9).toFixed(9)`). Previously these were stored as raw Numbers and finalized by `decimalizeMember` (the `each(this.price.base.addons, decimalizeMember, …)` calls), which rounds to 2 places — but `decimalizeMember` **skips strings**. Net effect: every add-on's public price flips format, e.g. a normal `snarf` add-on at `1.0` now reports:

```js
price.base.addons.snarf === "1.000000000"   // was "1.00"
```

This is a public-API contract change for every integrator reading `price.base.addons` / `price.addons`, beyond the sub-cent scope of this ticket. There's no existing test pinning the normal-add-on format, so CI stays green while the output silently changes.

### Suggested fix

Only emit the full-precision string when rounding to 2 places would actually change the value (sub-cent / >2-decimal prices); otherwise leave it as a Number so the existing `decimalizeMember` pass produces the normal `"1.00"` shape:

```js
      } else {
        const currencyPrice = addon.price[this.items.currency];
        unitPrice = currencyPrice.unit_amount_decimal != null
          ? parseFloat(currencyPrice.unit_amount_decimal)
          : currencyPrice.unit_amount;
        totalPrice = unitPrice * selectedQuantity;
      }

      // Decimal add-ons can carry more than 2 places of precision. Emit the
      // full-precision string only when rounding to 2 places would change the
      // value, so decimalizeMember's 2-place pass can't truncate it to "0.00".
      // Normal add-ons stay Numbers and get the standard 2-place treatment,
      // preserving the existing "1.00" output.
      const preservePrecision = value => round(value, 9) !== round(value, 2);
      this.price.base.addons[addon.code] = preservePrecision(unitPrice)
        ? decimalize(unitPrice, 9)
        : unitPrice;
      this.price.addons[addon.code] = preservePrecision(totalPrice) // DEPRECATED
        ? decimalize(totalPrice, 9)
        : totalPrice;
```

`round` and `decimalize` are both already imported, so no new imports. Normal add-ons round equal at 9 and 2 places → stay Numbers → `decimalizeMember` → `"1.00"` (original contract). Sub-cent values differ → `decimalize(_, 9)` → `"0.001234000"`, which `decimalizeMember` skips — so the existing sub-cent assertions still pass. `round()` also absorbs float noise (`0.1 + 0.2` rounds equal at both → `"0.30"`).

### Guard test

The regression slipped through because nothing pins the normal-add-on format. Adding this to `test/unit/pricing/subscription/subscription.test.js` locks the contract:

```js
it('emits standard 2-decimal output for a normal add-on', function (done) {
  this.pricing
    .plan('basic', { quantity: 1 })
    .addon('snarf', { quantity: 1 })
    .done(price => {
      assert.equal(price.base.addons['snarf'], '1.00');
      assert.equal(price.addons['snarf'], '1.00');
      done();
    });
});
```

Patrick-Duvall (approved):

Patrick-Duvall (approved):

PR #993, opened 2026-08-18
fix: Round percentage coupon discounts in integer cents to match API

McTraque:
## Description

Percentage-coupon discounts were computed by multiplying **dollar floats** (e.g. `34.90 * 0.15 = 5.234999…`), which drifts just below the half-cent tie and rounds **down** — so Checkout and Recurly.js displayed a discount one cent smaller than the API/invoice actually charged. On a $34.90 plan with a 15%-off coupon, Checkout showed a **$5.23** discount ($29.67 total) while the invoice charged a **$5.24** discount ($29.66).

The API is correct: it multiplies **integer cents** (`3490 * 0.15 = 523.5`) and rounds to `524`. This PR adds a `roundRateDiscount(amount, rate)` helper that reduces the discountable amount to integer cents **before** applying the rate — matching the API — and routes both discount paths through it:

- `checkout/calculations.js#discountAmounts()` — what the hosted Checkout / `CheckoutPricing` renders
- `subscription/calculations.js#discount()` — `SubscriptionPricing`, used by Recurly.js forms

Notes for reviewers:
- A prior attempt (`ac86950`, shipped in **v4.40.0**) standardized the rounding helper but kept multiplying dollars, so it only fixed cases that didn't drift below the tie. This supersedes it.
- The same dollar-float × rate pattern still exists for **tax**, which recurly.js intentionally treats as an estimate (the authoritative figure comes from the tax service). Left as-is, out of scope.

## Testing

**Automated Test Coverage:**
- Run: `make test-unit`
- Result: **1173 examples, 0 failures** — includes new regression tests for both pricing paths (`test/unit/pricing/checkout/checkout.test.js`, `test/unit/pricing/subscription/subscription.test.js`)
- Lint: `make lint` — clean

**Manual verification:**
1. Configure a plan priced `34.90` and a percentage coupon at `15%` off.
2. Build a `CheckoutPricing` (or `SubscriptionPricing`) with that plan and coupon and reprice.
3. Expected: `price.now.discount === '5.24'` (previously `'5.23'`), matching the invoice/API.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/lib/recurly/pricing/subscription/calculations.js
+++ b/lib/recurly/pricing/subscription/calculations.js
@@ -248,8 +248,8 @@ export default class Calculations {
     if (!coupon) return;
 
     if (coupon.discount.rate) {
-      var discountNow = round(this.price.now.subtotal * coupon.discount.rate);
-      var discountNext = round(this.price.next.subtotal * coupon.discount.rate);
+      var discountNow = round(this.price.now.subtotal * coupon.discount.rate, 6);
+      var discountNext = round(this.price.next.subtotal * coupon.discount.rate, 6);
       this.price.now.discount = discountNow;
       this.price.next.discount = discountNext;
     } else if (coupon.discount.type === 'free_trial') {
--- a/packages/public-api-fixture-server/fixtures/plans/pct-tie-plan.json
+++ b/packages/public-api-fixture-server/fixtures/plans/pct-tie-plan.json
@@ -1,0 +1,17 @@
+{
+  "code": "pct-tie-plan",
+  "name": "Pct Tie Plan",
+  "period": {
+    "interval": "months",
+    "length": 1
+  },
+  "price": {
+    "USD": {
+      "unit_amount": 34.90,
+      "symbol": "$",
+      "setup_fee": 0
+    }
+  },
+  "addons": [],
+  "tax_exempt": false
+}
--- a/test/unit/pricing/checkout/checkout.test.js
+++ b/test/unit/pricing/checkout/checkout.test.js
@@ -872,6 +872,18 @@ describe('CheckoutPricing', function () {
           });
       });
 
+      it('rounds a half-cent rate discount up to the nearest cent, matching the API', function (done) {
+        this.pricing
+          .adjustment({ amount: 34.90 })
+          .coupon('coop-pct-adjustments')
+          .reprice()
+          .done(price => {
+            assert.equal(price.now.discount, 5.24);
+            assert.equal(price.now.adjustments, 34.90);
+            done();
+          });
+      });
+
       describe('given a CheckoutPricing containing multiple subscriptions and adjustments', () => {
         beforeEach(function (done) {
           subscriptionPricingFactory('basic', this.recurly, sub => {
--- a/test/unit/pricing/subscription/subscription.test.js
+++ b/test/unit/pricing/subscription/subscription.test.js
@@ -697,6 +697,20 @@ describe('Recurly.Pricing.Subscription', function () {
         });
     });
 
+    it('rounds a half-cent rate discount up to the nearest cent, matching the API', function (done) {
+      this.pricing
+        .plan('pct-tie-plan', { quantity: 1 })
+        .address({
+          country: 'US',
+          postal_code: 'NoTax'
+        })
+        .coupon('coop-pct-all')
+        .done(function (price) {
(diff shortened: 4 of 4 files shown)


lib/util/decimalize.js
```diff
@@ -17,6 +17,22 @@ export function round (number, digits = 2) {
...
+ * is applied, so a half-cent result rounds to the nearest cent instead of
+ * drifting down through floating point dollar math. For example 15% of 34.90 is
+ * 5.24, not 5.23 (34.90 * 0.15 evaluates to 5.234999… as a dollar float).
+ *
+ * @param {Number} amount discountable amount in the major unit (e.g. dollars)
+ * @param {Number} rate discount rate (e.g. 0.15)
+ * @return {Number} discount in the major unit, rounded to the nearest cent
+ */
+export function roundRateDiscount (amount, rate) {
+  const cents = Math.round(amount * 100) * rate;
```
cbarton: I think this should use the `round` helper here well instead of the Math library to future proof for decimal based pricing.

lib/recurly/pricing/checkout/calculations.js
```diff
@@ -336,10 +336,10 @@ export default class Calculations {
         // Amounts are left zero
       } else if (coupon.discount.rate) {
         const { discountableNow, discountableNext } = this.discountableSubtotals(coupon, { setupFees: false });
-        discountNow = round(discountableNow * coupon.discount.rate, 6);
+        discountNow = roundRateDiscount(discountableNow, coupon.discount.rate);
```
cbarton: do these callsites need the 6 decimal precision?

lib/util/decimalize.js
```diff
@@ -17,6 +17,22 @@ export function round (number, digits = 2) {
...
+ * API does: the discountable amount is reduced to integer cents before the rate
+ * is applied, so a half-cent result rounds to the nearest cent instead of
+ * drifting down through floating point dollar math. For example 15% of 34.90 is
+ * 5.24, not 5.23 (34.90 * 0.15 evaluates to 5.234999… as a dollar float).
+ *
+ * @param {Number} amount discountable amount in the major unit (e.g. dollars)
+ * @param {Number} rate discount rate (e.g. 0.15)
+ * @return {Number} discount in the major unit, rounded to the nearest cent
+ */
+export function roundRateDiscount (amount, rate) {
```
cbarton: I think we need to accept a `digits` parameter here to mirror the signature for `round` so that the callsite in checkout calculations can pass in 6, if needed.

lib/util/decimalize.js
McTraque: Good call — I dropped the custom helper entirely (no more `Math` library). Digging in, the root cause was narrower than my first pass: on current master the checkout path already computes correctly (its `round(x, 6)` preserves `5.235`, which the final `decimalize` rounds to `5.24`). The only still-buggy path was `SubscriptionPricing`, which rounded straight to cents (`round(x)` = 2 digits). I fixed it to mirror the checkout path with `round(x, 6)`, so the intermediate keeps enough precision (incl. decimal pricing) and the final `decimalize` lands on the right cent. `decimalize.js` is now untouched.

lib/recurly/pricing/checkout/calculations.js
McTraque: Yes — and that was the key insight, thanks. The 6-digit precision here is load-bearing: it's exactly what makes the checkout path already correct on master (`round(5.2349…, 6)` = `5.235` → `decimalize` → `5.24`) and what preserves precision for decimal-based pricing. I reverted my change to this file entirely — this callsite is unchanged. The real bug was that `SubscriptionPricing` did *not* round to 6 (it used `round(x)` = 2 digits), so I brought it in line with this path.

lib/util/decimalize.js
McTraque: No longer needed — I removed the new helper. Instead of a `roundRateDiscount(amount, rate, digits)`, both paths now use the existing `round(x, 6)` directly: checkout already did, and I changed `SubscriptionPricing` to match. So there's no new signature to keep in sync with `round`.

cbarton (approved):

Patrick-Duvall (approved):

